PR #12, opened 2016-12-17
Restructure

alexd2580:
A branch for refactoring and restructuring. A lot of fixes have been introduced in the `create-bih` branch, which are not exclusive to BIHs.

### DIFF
--- a/src/bounding_hierarchy.rs
+++ b/src/bounding_hierarchy.rs
@@ -1,0 +1,36 @@
+//! This module defines the `BoundingHierarchy` trait.
+
+use aabb::Bounded;
+use ray::Ray;
+
+/// This trait defines an acceleration structure with space partitioning.
+/// This structure is used to efficiently compute ray-scene intersections.
+pub trait BoundingHierarchy {
+    /// Creates a new [`BoundingHierarchy`] from the `shapes` slice.
+    ///
+    /// [`BoundingHierarchy`]: trait.BoundingHierarchy.html
+    ///
+    /// # Examples
+    ///
+    /// // TODO example
+    ///
+    fn build<T: Bounded>(shapes: &[T]) -> Self;
+
+    /// Traverses the [`BoundingHierarchy`] recursively.
+    /// Returns a subset of `shapes`, in which the [`AABB`]s of the elements were hit by `ray`.
+    ///
+    /// [`BoundingHierarchy`]: trait.BoundingHierarchy.html
+    /// [`AABB`]: ../aabb/struct.AABB.html
+    ///
+    /// # Examples
+    ///
+    /// // TODO example
+    ///
+    fn traverse<'a, T: Bounded>(&'a self, ray: &Ray, shapes: &'a [T]) -> Vec<&T>;
+
+    /// Prints the [`BoundingHierarchy`] in a tree-like visualization.
+    ///
+    /// [`BoundingHierarchy`]: ../bounding_hierarchy/trait.BoundingHierarchy.html
+    ///
+    fn pretty_print(&self) {}
+}
--- a/src/bounding_hierarchy.rs
+++ b/src/bounding_hierarchy.rs
@@ -8,23 +8,72 @@ use ray::Ray;
 pub trait BoundingHierarchy {
     /// Creates a new [`BoundingHierarchy`] from the `shapes` slice.
     ///
-    /// [`BoundingHierarchy`]: trait.BoundingHierarchy.html
-    ///
     /// # Examples
     ///
-    /// // TODO example
+    /// ```
+    /// use bvh::aabb::{AABB, Bounded};
+    /// use bvh::bounding_hierarchy::BoundingHierarchy;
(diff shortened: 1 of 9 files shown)


src/testbase.rs
```diff
@@ -0,0 +1,300 @@
...
+}
+
+/// Given a ray, a bounding hierarchy, the complete list of shapes in the scene and a list of
+/// expected hits, verifies, whether the ray hits only the expected shapes.
+fn traverse_and_verify<BH: BoundingHierarchy>(ray_origin: Point3<f32>,
+                                              ray_direction: Vector3<f32>,
+                                              all_shapes: &Vec<UnitBox>,
+                                              bh: &BH,
+                                              expected_shapes: &HashSet<i32>) {
+    let ray = Ray::new(ray_origin, ray_direction);
```
FlaShG: I think it would make more sense to pass the ray instead of passing its components.

src/testbase.rs
```diff
@@ -0,0 +1,300 @@
...
+fn splitmix64(x: &mut u64) -> u64 {
+    *x = x.wrapping_add(0x9E3779B97F4A7C15u64);
+    let mut z = *x;
+    z = (z ^ (z >> 30)).wrapping_mul(0xBF58476D1CE4E5B9u64);
+    z = (z ^ (z >> 27)).wrapping_mul(0x94D049BB133111EBu64);
+    z ^ (z >> 31)
+}
+
+/// Generates a new Point3, mutates the seed.
+fn next_point3(seed: &mut u64) -> Point3<f32> {
```
FlaShG: The create_bih branch has a [different implementation](https://github.com/svenstaro/bvh/blob/create-bih/src/testbase.rs#L127). Which one is better?

alexd2580: This one is better, the random is not broken here.

src/testbase.rs
alexd2580: It does not make any semantic difference, but with this interface we can reduce a little more duplicate code (which is the only purpose of exporting the common functionality as a function).

src/bounding_hierarchy.rs
```diff
@@ -0,0 +1,36 @@
...
+/// This trait defines an acceleration structure with space partitioning.
+/// This structure is used to efficiently compute ray-scene intersections.
+pub trait BoundingHierarchy {
+    /// Creates a new [`BoundingHierarchy`] from the `shapes` slice.
+    ///
+    /// [`BoundingHierarchy`]: trait.BoundingHierarchy.html
+    ///
+    /// # Examples
+    ///
+    /// // TODO example
```
svenstaro: Fix the TODO

src/bounding_hierarchy.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    /// Traverses the [`BoundingHierarchy`] recursively.
+    /// Returns a subset of `shapes`, in which the [`AABB`]s of the elements were hit by `ray`.
+    ///
+    /// [`BoundingHierarchy`]: trait.BoundingHierarchy.html
+    /// [`AABB`]: ../aabb/struct.AABB.html
+    ///
+    /// # Examples
+    ///
+    /// // TODO example
```
svenstaro: Fix the TODO

src/flat_bvh.rs
```diff
@@ -260,130 +106,65 @@ impl BVHNode {
...
-    /// let bvh = BVH::build(&shapes);
-    /// let flat_bvh = bvh.flatten();
-    /// ```
-    pub fn flatten(&self) -> Vec<FlatNode> {
-        let mut vec = Vec::new();
-        self.root.flatten(&mut vec, 0);
-        vec
-    }
+/// A flat [`BVH`]. Represented by a vector of [`FlatNode`]s. The [`FlatBVH`] is designed for use
+/// where a recursive traversal of a data structure is not possible.
```
svenstaro: Add an example as to why this might be the case. IE "For instance on some architectures or devices (like GPUs) ..."

src/testbase.rs
```diff
@@ -0,0 +1,300 @@
...
+    });
+}
+
+/// Benchmark the traversal of a `BoundingHierarchy` with `n` triangles.
+pub fn intersect_n_triangles<T: BoundingHierarchy>(n: u64, b: &mut ::test::Bencher) {
+    let triangles = create_n_cubes(n);
+    let structure = T::build(&triangles);
+    let mut seed = 0;
+
+    // if n <= 50 {
```
svenstaro: Why this commented code?

src/bounding_hierarchy.rs
alexd2580: Added an example 5db8335.

src/bounding_hierarchy.rs
alexd2580: Added an example 5db8335.

src/testbase.rs
alexd2580: Removed in 5db8335.

PR #14, opened 2017-01-19
WIP: Dynamic bvh optimization

FlaShG:
The optimization extension is in a working state.
We're missing tests and benchmarks, but the code itself could use an early review.

This pull request is WIP and should not be merged before tests and benchmarks are implemented.

### DIFF
--- a/src/aabb.rs
+++ b/src/aabb.rs
@@ -113,9 +113,9 @@ impl AABB {
     /// use bvh::aabb::AABB;
     /// use bvh::nalgebra::Point3;
     ///
-    /// let aabb = AABB::with_bounds(Point3::new(-1.0,-1.0,-1.0), Point3::new(1.0,1.0,1.0));
-    /// let point_inside = Point3::new(0.125,-0.25,0.5);
-    /// let point_outside = Point3::new(1.0,-2.0,4.0);
+    /// let aabb = AABB::with_bounds(Point3::new(-1.0, -1.0, -1.0), Point3::new(1.0, 1.0, 1.0));
+    /// let point_inside = Point3::new(0.125, -0.25, 0.5);
+    /// let point_outside = Point3::new(1.0, -2.0, 4.0);
     ///
     /// assert!(aabb.contains(&point_inside));
     /// assert!(!aabb.contains(&point_outside));
@@ -138,11 +138,11 @@ impl AABB {
     /// use bvh::aabb::AABB;
     /// use bvh::nalgebra::Point3;
     ///
-    /// let aabb = AABB::with_bounds(Point3::new(-1.0,-1.0,-1.0), Point3::new(1.0,1.0,1.0));
-    /// let point_barely_inside = Point3::new(1.0000001,-1.0000001,1.000000001);
-    /// let point_outside = Point3::new(1.0,-2.0,4.0);
+    /// let aabb = AABB::with_bounds(Point3::new(-1.0, -1.0, -1.0), Point3::new(1.0, 1.0, 1.0));
+    /// let point_barely_outside = Point3::new(1.000_000_1, -1.000_000_1, 1.000_000_001);
+    /// let point_outside = Point3::new(1.0, -2.0, 4.0);
     ///
-    /// assert!(aabb.approx_contains_eps(&point_barely_inside, EPSILON));
+    /// assert!(aabb.approx_contains_eps(&point_barely_outside, EPSILON));
     /// assert!(!aabb.approx_contains_eps(&point_outside, EPSILON));
     /// ```
     ///
@@ -165,9 +165,9 @@ impl AABB {
     /// use bvh::nalgebra::Point3;
     ///
     /// let aabb = AABB::with_bounds(Point3::new(-1.0, -1.0, -1.0), Point3::new(1.0, 1.0, 1.0));
-    /// let point_barely_inside = Point3::new(1.00_000_01, 1.00_000_01, 1.00_000_01);
+    /// let point_barely_outside = Point3::new(1.000_000_1, 1.000_000_1, 1.000_000_1);
     /// let center = aabb.center();
-    /// let inner_aabb = AABB::with_bounds(center, point_barely_inside);
+    /// let inner_aabb = AABB::with_bounds(center, point_barely_outside);
     ///
     /// assert!(aabb.approx_contains_aabb_eps(&inner_aabb, EPSILON));
     /// ```
@@ -188,9 +188,9 @@ impl AABB {
     /// use bvh::nalgebra::Point3;
     ///
     /// let aabb = AABB::with_bounds(Point3::new(-1.0, -1.0, -1.0), Point3::new(1.0, 1.0, 1.0));
-    /// let point_barely_inside_min = Point3::new(-1.00_000_01, -1.00_000_01, -1.00_000_01);
-    /// let point_barely_inside_max = Point3::new(1.00_000_01, 1.00_000_01, 1.00_000_01);
-    /// let other = AABB::with_bounds(point_barely_inside_min, point_barely_inside_max);
+    /// let point_barely_outside_min = Point3::new(-1.000_000_1, -1.000_000_1, -1.000_000_1);
+    /// let point_barely_outside_max = Point3::new(1.000_000_1, 1.000_000_1, 1.000_000_1);
+    /// let other = AABB::with_bounds(point_barely_outside_min, point_barely_outside_max);
     ///
     /// assert!(aabb.relative_eq(&other, EPSILON));
     /// ```
@@ -209,13 +209,13 @@ impl AABB {
     /// use bvh::aabb::AABB;
     /// use bvh::nalgebra::Point3;
     ///
-    /// let aabb1 = AABB::with_bounds(Point3::new(-101.0,0.0,0.0), Point3::new(-100.0,1.0,1.0));
-    /// let aabb2 = AABB::with_bounds(Point3::new(100.0,0.0,0.0), Point3::new(101.0,1.0,1.0));
+    /// let aabb1 = AABB::with_bounds(Point3::new(-101.0, 0.0, 0.0), Point3::new(-100.0, 1.0, 1.0));
+    /// let aabb2 = AABB::with_bounds(Point3::new(100.0, 0.0, 0.0), Point3::new(101.0, 1.0, 1.0));
     /// let joint = aabb1.join(&aabb2);
     ///
-    /// let point_inside_aabb1 = Point3::new(-100.5,0.5,0.5);
-    /// let point_inside_aabb2 = Point3::new(100.5,0.5,0.5);
-    /// let point_inside_joint = Point3::new(0.0,0.5,0.5);
+    /// let point_inside_aabb1 = Point3::new(-100.5, 0.5, 0.5);
+    /// let point_inside_aabb2 = Point3::new(100.5, 0.5, 0.5);
+    /// let point_inside_joint = Point3::new(0.0, 0.5, 0.5);
     ///
     /// # assert!(aabb1.contains(&point_inside_aabb1));
     /// # assert!(!aabb1.contains(&point_inside_aabb2));
@@ -249,9 +249,9 @@ impl AABB {
     /// use bvh::aabb::AABB;
     /// use bvh::nalgebra::Point3;
     ///
-    /// let point1 = Point3::new(0.0,0.0,0.0);
-    /// let point2 = Point3::new(1.0,1.0,1.0);
-    /// let point3 = Point3::new(2.0,2.0,2.0);
+    /// let point1 = Point3::new(0.0, 0.0, 0.0);
+    /// let point2 = Point3::new(1.0, 1.0, 1.0);
+    /// let point3 = Point3::new(2.0, 2.0, 2.0);
     ///
     /// let aabb = AABB::empty();
     /// assert!(!aabb.contains(&point1));
--- a/src/aabb.rs
+++ b/src/aabb.rs
@@ -155,6 +155,52 @@ impl AABB {
         (p.z - self.min.z) > -epsilon && (p.z - self.max.z) < epsilon
     }
 
+    /// Returns true if the `other` [`AABB`] is approximately inside this [`AABB`]
+    /// with respect to some `epsilon`.
+    ///
+    /// # Examples
+    /// ```
+    /// use bvh::EPSILON;
+    /// use bvh::aabb::AABB;
+    /// use bvh::nalgebra::Point3;
+    ///
+    /// let aabb = AABB::with_bounds(Point3::new(-1.0, -1.0, -1.0), Point3::new(1.0, 1.0, 1.0));
+    /// let point_barely_inside = Point3::new(1.00_000_01, 1.00_000_01, 1.00_000_01);
+    /// let center = aabb.center();
+    /// let inner_aabb = AABB::with_bounds(center, point_barely_inside);
+    ///
+    /// assert!(aabb.approx_contains_aabb_eps(&inner_aabb, EPSILON));
+    /// ```
+    ///
+    /// [`AABB`]: struct.AABB.html
+    pub fn approx_contains_aabb_eps(&self, other: &AABB, epsilon: f32) -> bool {
+        self.approx_contains_eps(&other.min, epsilon) &&
+        self.approx_contains_eps(&other.max, epsilon)
+    }
+
+    /// Returns true if the `other` [`AABB`] is approximately equal to this [`AABB`]
+    /// with respect to some `epsilon`.
+    ///
+    /// # Examples
+    /// ```
+    /// use bvh::EPSILON;
+    /// use bvh::aabb::AABB;
+    /// use bvh::nalgebra::Point3;
+    ///
+    /// let aabb = AABB::with_bounds(Point3::new(-1.0, -1.0, -1.0), Point3::new(1.0, 1.0, 1.0));
+    /// let point_barely_inside_min = Point3::new(-1.00_000_01, -1.00_000_01, -1.00_000_01);
+    /// let point_barely_inside_max = Point3::new(1.00_000_01, 1.00_000_01, 1.00_000_01);
+    /// let other = AABB::with_bounds(point_barely_inside_min, point_barely_inside_max);
+    ///
+    /// assert!(aabb.relative_eq(&other, EPSILON));
+    /// ```
+    ///
+    /// [`AABB`]: struct.AABB.html
+    pub fn relative_eq(&self, other: &AABB, epsilon: f32) -> bool {
+        relative_eq!(self.min, other.min, epsilon = epsilon) &&
+        relative_eq!(self.max, other.max, epsilon = epsilon)
+    }
+
     /// Returns a new minimal [`AABB`] which contains both this [`AABB`] and `other`.
     /// The result is the convex hull of the both [`AABB`]s.
     ///
--- a/src/bounding_hierarchy.rs
+++ b/src/bounding_hierarchy.rs
@@ -147,7 +147,7 @@ pub trait BoundingHierarchy {
 
     /// Prints the [`BoundingHierarchy`] in a tree-like visualization.
     ///
-    /// [`BoundingHierarchy`]: ../bounding_hierarchy/trait.BoundingHierarchy.html
+    /// [`BoundingHierarchy`]: trait.BoundingHierarchy.html
     ///
     fn pretty_print(&self) {}
 }
--- a/src/bounding_hierarchy.rs
+++ b/src/bounding_hierarchy.rs
@@ -49,7 +49,7 @@ pub trait BoundingHierarchy {
     ///
     fn build<T: BHShape>(shapes: &mut [T]) -> Self;
 
-    /// Traverses the [`BoundingHierarchy`] recursively.
+    /// Traverses the [`BoundingHierarchy`].
     /// Returns a subset of `shapes`, in which the [`AABB`]s of the elements were hit by `ray`.
     ///
     /// # Examples
--- a/src/bounding_hierarchy.rs
+++ b/src/bounding_hierarchy.rs
@@ -15,23 +15,50 @@ pub trait BoundingHierarchy {
     /// use bvh::bounding_hierarchy::BoundingHierarchy;
     /// use bvh::nalgebra::{Point3, Vector3};
     /// # use bvh::bounding_hierarchy::BHShape;
-    /// #
-    /// # impl BHShape for AABB {
-    /// #     fn set_bh_node_index(&mut self, index: usize) { }
-    /// #     fn bh_node_index(&self) -> usize { 0 }
+    /// # pub struct UnitBox {
+    /// #     pub id: i32,
+    /// #     pub pos: Point3<f32>,
+    /// #     node_index: usize,
     /// # }
     /// #
-    /// # fn create_bounded_shapes() -> Vec<AABB> {
+    /// # impl UnitBox {
+    /// #     pub fn new(id: i32, pos: Point3<f32>) -> UnitBox {
+    /// #         UnitBox {
+    /// #             id: id,
+    /// #             pos: pos,
+    /// #             node_index: 0,
+    /// #         }
+    /// #     }
+    /// # }
+    /// #
+    /// # impl Bounded for UnitBox {
+    /// #     fn aabb(&self) -> AABB {
+    /// #         let min = self.pos + Vector3::new(-0.5, -0.5, -0.5);
+    /// #         let max = self.pos + Vector3::new(0.5, 0.5, 0.5);
+    /// #         AABB::with_bounds(min, max)
+    /// #     }
+    /// # }
+    /// #
+    /// # impl BHShape for UnitBox {
+    /// #     fn set_bh_node_index(&mut self, index: usize) {
+    /// #         self.node_index = index;
+    /// #     }
+    /// #
+    /// #     fn bh_node_index(&self) -> usize {
+    /// #         self.node_index
+    /// #     }
+    /// # }
+    /// #
+    /// # fn create_bhshapes() -> Vec<UnitBox> {
     /// #     let mut shapes = Vec::new();
-    /// #     let offset = Vector3::new(1.0, 1.0, 1.0);
-    /// #     for i in 0..1000u32 {
+    /// #     for i in 0..1000 {
     /// #         let position = Point3::new(i as f32, i as f32, i as f32);
-    /// #         shapes.push(AABB::with_bounds(position - offset, position + offset));
+    /// #         shapes.push(UnitBox::new(i, position));
     /// #     }
     /// #     shapes
     /// # }
     ///
-    /// let mut shapes = create_bounded_shapes();
+    /// let mut shapes = create_bhshapes();
     /// // Construct a normal `BVH`.
     /// {
     ///     use bvh::bvh::BVH;
@@ -41,7 +68,7 @@ pub trait BoundingHierarchy {
     /// // Or construct a `FlatBVH`.
     /// {
     ///     use bvh::flat_bvh::FlatBVH;
-    ///     let bvh = FlatBVH::build(&shapes);
+    ///     let bvh = FlatBVH::build(&mut shapes);
     /// }
     /// ```
     ///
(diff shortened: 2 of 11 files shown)


src/bounding_hierarchy.rs
```diff
@@ -14,6 +14,13 @@ pub trait BoundingHierarchy {
     /// use bvh::aabb::{AABB, Bounded};
     /// use bvh::bounding_hierarchy::BoundingHierarchy;
     /// use bvh::nalgebra::{Point3, Vector3};
+    /// # use bvh::bounding_hierarchy::BHShape;
+    /// #
+    /// # impl BHShape for AABB {
```
alexd2580: This `impl` is invalid. Unfortunately we need to define an explicit structure and define BVHShape for it.
And also unfortunately we can't hide or deduplicate this boilerplate code using utilities or macros, because doctest are executed in an integration test environment.

src/bvh/bvh.rs
```diff
@@ -22,34 +20,55 @@ use ray::Ray;
...
+#[derive(Debug, Copy, Clone)]
 pub enum BVHNode {
     /// Leaf node.
     Leaf {
-        /// The shapes contained in this leaf.
-        shapes: Vec<usize>,
+        /// The node's parent.
+        parent: usize,
+
+        /// The node's depth
```
alexd2580: Full stop at the end of the sentence.

src/bvh/bvh.rs
```diff
@@ -22,34 +20,55 @@ use ray::Ray;
...
+
+        /// The shape contained in this leaf.
+        shape: usize,
     },
     /// Inner node.
     Node {
+        /// The node's parent.
+        parent: usize,
+
+        /// The node's depth
```
alexd2580: See above.

src/bvh/bvh.rs
```diff
@@ -22,34 +20,55 @@ use ray::Ray;
...
+        depth: u32,
+
         /// The convex hull of the shapes' `AABB`s in child_l.
         child_l_aabb: AABB,
 
         /// Left subtree.
```
alexd2580: The comments are a little outdated, this is not the subtree itself, it's the index of the subtree.

src/bvh/bvh.rs
```diff
@@ -22,34 +20,55 @@ use ray::Ray;
...
-        child_r: Box<BVHNode>,
+        child_r: usize,
     },
+    /// Dummy node for later replacement
+    Dummy,
 }
 
 impl BVHNode {
     /// Builds a [`BVHNode`] recursively using SAH partitioning.
+    /// Returns the index of the node in the nodes vector.
```
alexd2580: Add documentation on the input arguments.

src/bvh/bvh.rs
```diff
@@ -113,22 +137,37 @@ impl BVHNode {
...
-            let shape_center = shape_aabb.center();
-
-            // Get the relative position of the shape centroid [0.0..1.0]
-            let bucket_num_relative = (shape_center[split_axis] - centroid_bounds.min[split_axis]) /
-                                      split_axis_size;
+        if split_axis_size < EPSILON {
+            // Spread the remaining shapes evenly across buckets,
+            // instead of using a heuristic
+            let mut bucket_num = 0;
+            for idx in &indices {
```
alexd2580: You can use enumerate to get the index of `idx` in `&indIces` and use it to determine the bucket number
instead of using the mutable `bucket_num` approach.

src/bvh/bvh.rs
```diff
@@ -163,30 +202,44 @@ impl BVHNode {
...
+
+        node_index
     }
 
     /// Prints a textual representation of the recursive [`BVH`] structure.
     ///
     /// [`BVH`]: struct.BVH.html
     ///
-    fn pretty_print(&self, depth: usize) {
+    fn pretty_print(&self, nodes: &Vec<BVHNode>, depth: usize) {
```
alexd2580: This method no longer conceptually fits into the implementation of `BVHNode`. It requires the whole tree, instead of just the node itself. It would make more sense to move this to `BVH`, what do you think?)

src/bvh/bvh.rs
```diff
@@ -237,8 +291,14 @@ impl BVH {
     ///
     /// ```
     /// use bvh::aabb::{AABB, Bounded};
-    /// use bvh::bvh::BVH;
+    /// use bvh::bvh::{BVH, BHShape};
+    /// use
```
alexd2580: Hm

src/bvh/bvh.rs
FlaShG: I agree, will move it.

src/bvh/bvh.rs
```diff
@@ -197,20 +253,18 @@ impl BVHNode {
     /// [`BVH`]: struct.BVH.html
     /// [`Vec`]: https://doc.rust-lang.org/std/vec/struct.Vec.html
     ///
-    pub fn traverse_recursive(&self, ray: &Ray, indices: &mut Vec<usize>) {
+    pub fn traverse_recursive(&self, nodes: &Vec<BVHNode>, ray: &Ray, indices: &mut Vec<usize>) {
```
alexd2580: Modify this function, so that it is static (no &self) and instead takes an index of the current node.

It looks like an antipattern to call a function on an element of an array and pass the array to the function.

src/bvh/bvh.rs
```diff
@@ -249,21 +309,23 @@ impl BVH {
...
 }
 
 impl BoundingHierarchy for BVH {
-    fn build<T: Bounded>(shapes: &[T]) -> BVH {
+    fn build<T: BHShape>(shapes: &mut [T]) -> BVH {
         let indices = (0..shapes.len()).collect::<Vec<usize>>();
-        let root = BVHNode::build(shapes, indices);
-        BVH { root: root }
+        let mut nodes = Vec::new();
+        BVHNode::build(shapes, indices, &mut nodes, 0, 0);
```
alexd2580: Why is the return vallue irrelevant here?

src/bvh/bvh.rs
```diff
@@ -299,13 +366,13 @@ impl BoundingHierarchy for BVH {
     /// ```
     fn traverse<'a, T: Bounded>(&'a self, ray: &Ray, shapes: &'a [T]) -> Vec<&T> {
         let mut indices = Vec::new();
-        self.root.traverse_recursive(ray, &mut indices);
+        self.nodes[0].traverse_recursive(&self.nodes, ray, &mut indices);
```
alexd2580: This looks bad. In terms of rust and in terms of design. `traverse_recursive` is canned on an element of the `self.nodes` array, while the array itself is passed to the function.
Modify the `BVHNode::traverse_recursive` function, make it static, and pass the index of the current node.

src/bvh/bvh.rs
```diff
@@ -299,13 +366,13 @@ impl BoundingHierarchy for BVH {
...
         let mut hit_shapes = Vec::new();
         for index in &indices {
             let shape = &shapes[*index];
-            if ray.intersects_aabb(&shape.aabb()) {
-                hit_shapes.push(shape);
-            }
+            // if ray.intersects_aabb(&shape.aabb()) {
+            hit_shapes.push(shape);
```
alexd2580: why is the `if` commented out?

src/bvh/bvh.rs
```diff
@@ -315,7 +382,7 @@ impl BoundingHierarchy for BVH {
     /// [`BVH`]: struct.BVH.html
     ///
     fn pretty_print(&self) {
-        self.root.pretty_print(0);
+        self.nodes[0].pretty_print(&self.nodes, 0);
```
alexd2580: see above

src/bvh/bvh.rs
```diff
@@ -325,6 +392,7 @@ pub mod tests {
     use testbase::{build_some_bh, traverse_some_bh, build_1200_triangles_bh,
                    build_12k_triangles_bh, build_120k_triangles_bh, intersect_1200_triangles_bh,
                    intersect_12k_triangles_bh, intersect_120k_triangles_bh};
+    use std::collections::HashSet;
```
alexd2580: Is this include needed here?

src/bvh/mod.rs
```diff
@@ -0,0 +1,9 @@
+//! This module defines a [`BVH`].
+//!
+//! [`BVH`]: struct.BVH.html
+//!
+
+pub use self::bvh::*;
```
alexd2580: Put the mods before the uses.

src/bvh/optimization.rs
```diff
@@ -0,0 +1,457 @@
...
+                    max_depth = depth;
+                    sweep_node_indices.clear();
+                    sweep_node_indices.push(*refit_node_index);
+                } else if depth == max_depth {
+                    sweep_node_indices.push(*refit_node_index);
+                }
+            }
+
+            // Try to find a useful tree rotation with all nodes previously found
+            for sweep_node_index in sweep_node_indices.iter() {
```
alexd2580: You can immutably iterate over vectors using `&sweep_node_indices`.

src/bvh/optimization.rs
```diff
@@ -0,0 +1,457 @@
...
+
+    /// Checks if there is a way to rotate a child and a grandchild (or two grandchildren) of
+    /// the given node (specified by `node_index`) that would improve the `BVH`.
+    /// If there is, the best rotation found is performed.
+    /// Relies on the children and transitive children of the given node having correct AABBs.
+    ///
+    /// Returns Some(index_of_node) if a new node was found that should be used for optimization.
+    ///
+    fn try_rotate<Shape: BHShape>(&mut self, node_index: usize, shapes: &[Shape]) -> Option<usize> {
+        let mut nodes = &mut self.nodes;
```
alexd2580: Does `nodes` have to be mutable? It is assigned a mutable reference, but its content doesn't change, does it?

src/bvh/bvh.rs
FlaShG: The return value is the index of the node added. This index is needed for the recursive call within the function.
In this case, it just returns 0, and we don't care.

src/bvh/optimization.rs
```diff
@@ -0,0 +1,457 @@
...
+    }
+
+    /// Checks if there is a way to rotate a child and a grandchild (or two grandchildren) of
+    /// the given node (specified by `node_index`) that would improve the `BVH`.
+    /// If there is, the best rotation found is performed.
+    /// Relies on the children and transitive children of the given node having correct AABBs.
+    ///
+    /// Returns Some(index_of_node) if a new node was found that should be used for optimization.
+    ///
+    fn try_rotate<Shape: BHShape>(&mut self, node_index: usize, shapes: &[Shape]) -> Option<usize> {
```
alexd2580: Is it possible to split this function into smaller parts?

src/bvh/bvh.rs
FlaShG: I haven't been sure for some time about it.

I just removed the `if` completely now. The reason is that there is only one shape per leaf now, and if you get to the leaf during traversal, the shape's AABB has already been tested against.

src/bvh/bvh.rs
FlaShG: Fixed in 21a02e8.

src/bvh/bvh.rs
FlaShG: Good point, updated in 945a1b5.

src/bvh/bvh.rs
FlaShG: Updated in 933752a.

src/bvh/bvh.rs
FlaShG: Updated in 933752a.

src/bvh/bvh.rs
FlaShG: Updated in a00b8e1.

src/bvh/bvh.rs
FlaShG: Nope. Removed in 7f11253.

src/bvh/mod.rs
FlaShG: The [Rust Book](https://doc.rust-lang.org/book/crates-and-modules.html) does it this way:
```
pub use self::greetings::hello;
pub use self::farewells::goodbye;

mod greetings;
mod farewells;
```

src/testbase.rs
```diff
@@ -37,6 +39,21 @@ impl Bounded for UnitBox {
     }
 }
 
+impl BHShape for UnitBox {
+    fn set_bh_node_index(&mut self, index: usize) {
+        self.node_index = index;
+        if self.has_good_bh_position {
+            panic!("An UnitBox's BVHNode has been moved even though it shouldn't have.");
```
alexd2580: We need some structure which sopports moving, for future tests. Probably you can then just remove this check here.

src/bvh/optimization.rs
FlaShG: Changed in 34f39a8.

src/bvh/optimization.rs
FlaShG: It gets mutably borrowed in lines 101, 235 and 246.
We could borrow `nodes` mutably in these places and make the borrow immutable in line 71, but I don't see too much benefit in that.

src/bvh/optimization.rs
FlaShG: I don't see any way to do that so it brings an actual benefit. Any splits into additional scopes would just yield more code around it to keep the borrow checker happy.
I've been generously commenting its sections instead.

src/bounding_hierarchy.rs
```diff
@@ -75,11 +143,30 @@ pub trait BoundingHierarchy {
     /// [`BoundingHierarchy`]: trait.BoundingHierarchy.html
     /// [`AABB`]: ../aabb/struct.AABB.html
     ///
-    fn traverse<'a, T: Bounded>(&'a self, ray: &Ray, shapes: &'a [T]) -> Vec<&T>;
+    fn traverse<'a, Shape: BHShape>(&'a self, ray: &Ray, shapes: &'a [Shape]) -> Vec<&Shape>;
 
     /// Prints the [`BoundingHierarchy`] in a tree-like visualization.
     ///
     /// [`BoundingHierarchy`]: ../bounding_hierarchy/trait.BoundingHierarchy.html
```
alexd2580: Fix the link. `../bounding_hierarchy` is the current directory.
```
/// [`BoundingHierarchy`]: trait.BoundingHierarchy.html
```

src/bvh/bvh.rs
```diff
@@ -1,55 +1,79 @@
...
 use ray::Ray;
+use std::f32;
+use std::iter::repeat;
 
-/// Enum which describes the union type of a node in a [`BVH`].
-/// This structure does not allow for storing a root node's [`AABB`]. Therefore rays
-/// which do not hit the root [`AABB`] perform two [`AABB`] tests (left/right) instead of one.
-/// On the other hand this structure decreases the total number of indirections when traversing
-/// the BVH. Only those nodes are accessed, which are definetely hit.
+/// The BVHNode enum that describes a node in a [`BVH`].
```
alexd2580: Highlight and link `BVHNode`.

src/bvh/bvh.rs
```diff
@@ -1,55 +1,79 @@
...
+/// It's either a leaf node and references a shape (by holding its index)
+/// or a regular node that has two children nodes.
```
alexd2580: `[...] two child nodes.`

src/bvh/bvh.rs
```diff
@@ -1,55 +1,79 @@
...
 
-        /// Right subtree.
-        child_r: Box<BVHNode>,
+        /// Index of the right subtree's root node.
+        child_r: usize,
     },
 }
 
 impl BVHNode {
+    fn create_dummy() -> BVHNode {
```
alexd2580: Add a short docstring explaining, why we need this function.

src/bounding_hierarchy.rs
FlaShG: Fixed in cf4acf22f3366f7a6341d05451791f000aca846a.

src/bvh/bvh.rs
```diff
@@ -64,19 +88,24 @@ impl BVHNode {
...
-            return BVHNode::Leaf { shapes: indices };
+        // If there is only one element left, don't split anymore
+        if indices.len() == 1 {
+            let shape_index = indices[0];
+            nodes.push(BVHNode::Leaf {
+                parent: parent,
+                depth: depth,
+                shape: shape_index,
+            });
+            let node_index = nodes.len() - 1;
```
alexd2580: Move this between lines 93 and 94, and remove `-1`.

src/bvh/bvh.rs
FlaShG: Fixed in e0952351159e55954e24816bb0f1d3f80c4cfe68.

src/bvh/bvh.rs
FlaShG: Added in a9bd617567edc05c7f2e2d0be1ad748912c8fe7c.

src/bvh/bvh.rs
FlaShG: Changed in 520224d55a15884cd031f3e2c311851c0b69ea6e.

src/bvh/bvh.rs
```diff
@@ -113,22 +142,36 @@ impl BVHNode {
         let mut buckets = [Bucket::empty(); NUM_BUCKETS];
         let mut bucket_assignments: [Vec<usize>; NUM_BUCKETS] = Default::default();
```
alexd2580: I'm going to refactor this part in a moment.

src/flat_bvh.rs
```diff
@@ -136,12 +137,46 @@ impl BVH {
     /// use bvh::bvh::BVH;
     /// use bvh::nalgebra::{Point3, Vector3};
     /// use bvh::ray::Ray;
-    /// # fn create_bounded_shapes() -> Vec<AABB> {
+    /// # use bvh::bounding_hierarchy::BHShape;
+    /// # pub struct UnitBox {
```
Prior99: find a way to make this DRY. Let us not 6times copy'n'paste this stuff.

Prior99: Maybe move this into `examples/` directory and define a unitbox mod there.

src/bvh/mod.rs
alexd2580: Done.

src/aabb.rs
```diff
@@ -155,6 +155,52 @@ impl AABB {
...
+    /// with respect to some `epsilon`.
+    ///
+    /// # Examples
+    /// ```
+    /// use bvh::EPSILON;
+    /// use bvh::aabb::AABB;
+    /// use bvh::nalgebra::Point3;
+    ///
+    /// let aabb = AABB::with_bounds(Point3::new(-1.0, -1.0, -1.0), Point3::new(1.0, 1.0, 1.0));
+    /// let point_barely_inside = Point3::new(1.00_000_01, 1.00_000_01, 1.00_000_01);
```
alexd2580: Done

src/aabb.rs
```diff
@@ -155,6 +155,52 @@ impl AABB {
...
+    /// with respect to some `epsilon`.
+    ///
+    /// # Examples
+    /// ```
+    /// use bvh::EPSILON;
+    /// use bvh::aabb::AABB;
+    /// use bvh::nalgebra::Point3;
+    ///
+    /// let aabb = AABB::with_bounds(Point3::new(-1.0, -1.0, -1.0), Point3::new(1.0, 1.0, 1.0));
+    /// let point_barely_inside_min = Point3::new(-1.00_000_01, -1.00_000_01, -1.00_000_01);
```
alexd2580: Depends on how you interpret it. The point is mathematically barely outside, but the implementation classifies the point as inside - barely inside.

src/aabb.rs
```diff
@@ -155,6 +155,52 @@ impl AABB {
...
+    /// let point_barely_inside_max = Point3::new(1.00_000_01, 1.00_000_01, 1.00_000_01);
```
alexd2580: asd

alexd2580: See above

src/bounding_hierarchy.rs
```diff
@@ -53,14 +87,48 @@ pub trait BoundingHierarchy {
     /// use bvh::bvh::BVH;
     /// use bvh::nalgebra::{Point3, Vector3};
     /// use bvh::ray::Ray;
-    /// # fn create_bvh() -> (BVH, Vec<AABB>) {
+    /// # use bvh::bounding_hierarchy::BHShape;
+    /// # pub struct UnitBox {
```
alexd2580: You can't. We don't want to expose things from this lib which do not belong into the public interface.
And unitbox is very clearly just a testing thing.

src/aabb.rs
```diff
@@ -155,6 +155,52 @@ impl AABB {
...
+    /// let other = AABB::with_bounds(point_barely_inside_min, point_barely_inside_max);
+    ///
+    /// assert!(aabb.relative_eq(&other, EPSILON));
+    /// ```
+    ///
+    /// [`AABB`]: struct.AABB.html
+    pub fn relative_eq(&self, other: &AABB, epsilon: f32) -> bool {
```
alexd2580: This will stay this way, because the approx crate has the same naming scheme.
in approx, relative_eq is an imprecise floating point comparison.
Then we traverse nalgebra, which uses approx to define imprecise fpoint comparison on vectors with `relative_eq`.
Our AABBs adhere to the same comparison rules.

src/bvh/bvh.rs
```diff
@@ -0,0 +1,401 @@
...
+            Bucket {
+                size: a.size + b.size,
+                aabb: a.aabb.join(&b.aabb),
+            }
+        }
+
+        // Create six buckets, and six index assignment vectors
+        const NUM_BUCKETS: usize = 6;
+        let mut buckets = [Bucket::empty(); NUM_BUCKETS];
+        let mut bucket_assignments: [Vec<usize>; NUM_BUCKETS] = Default::default();
```
alexd2580: Here we compute the SAH weight of the different split options.
`buckets` stores the bucket definitions of the splits, while
`bucket_assignments` stores the shape indices which were put in this bucket.
This is stored so that we don't have to recompute the índex-partitioning.

src/bvh/bvh.rs
```diff
@@ -0,0 +1,401 @@
...
+            }
+        }
+
+        // Compute the costs for each configuration and
+        // select the configuration with the minimal costs
+        let mut min_bucket = 0;
+        let mut min_cost = f32::INFINITY;
+        let mut child_l_aabb = AABB::empty();
+        let mut child_r_aabb = AABB::empty();
+        for i in 0..(NUM_BUCKETS - 1) {
```
alexd2580: You could, BUT it would bring no merit at all, because you would have to recompute the split in line 187 anyway. You would have to fold on `(0..NUM_BUCKETS-1)` which can be expressed better with a loop.

src/bvh/bvh.rs
```diff
@@ -0,0 +1,401 @@
...
+
+        fn join_vectors(vectors: &mut [Vec<usize>]) -> Vec<usize> {
+            let mut result = Vec::new();
+            for mut vector in vectors.iter_mut() {
+                result.append(&mut vector);
+            }
+            result
+        }
+
+        // Join together all index buckets, and proceed recursively
```
alexd2580: A bucket (in SAH) defines a partition containing both the metadata, such as the union aabb, the cost, a.s.o. AND the shapes. We have split this into two containers. the one with the metadata and one with the indices.

src/bvh/bvh.rs
```diff
@@ -0,0 +1,401 @@
...
+    /// Returns a subset of `shapes`, in which the [`AABB`]s of the elements were hit by `ray`.
+    ///
+    /// [`BVH`]: struct.BVH.html
+    /// [`AABB`]: ../aabb/struct.AABB.html
+    ///
+    pub fn traverse<'a, Shape: Bounded>(&'a self, ray: &Ray, shapes: &'a [Shape]) -> Vec<&Shape> {
+        let mut indices = Vec::new();
+        BVHNode::traverse_recursive(&self.nodes, 0, ray, &mut indices);
+        let mut hit_shapes = Vec::new();
+        for index in &indices {
```
alexd2580: Yes.

src/bvh/optimization.rs
```diff
@@ -0,0 +1,1039 @@
...
+//! [`BVH`]: struct.BVH.html
+//!
+
+use bvh::*;
+use bounding_hierarchy::BHShape;
+use aabb::AABB;
+use std::collections::HashSet;
+use rand::{thread_rng, Rng};
+
+// TODO Consider: Instead of getting the scene's shapes passed, let leaf nodes store an AABB
```
alexd2580: Done.

src/bvh/optimization.rs
```diff
@@ -0,0 +1,1039 @@
...
+impl OptimizationIndex {
+    fn index(&self) -> usize {
+        match self {
+            &OptimizationIndex::Refit(index) |
+            &OptimizationIndex::FixAABBs(index) => index,
+        }
+    }
+}
+
+impl BVHNode {
```
alexd2580: I will try to move the utility-functions out of here.

src/bvh/optimization.rs
```diff
@@ -0,0 +1,1039 @@
...
+                                    shapes: &[Shape]) {
+        // `refit_node_indices` will contain the indices of the leaf nodes
+        // that reference the given shapes, sorted by their depth
+        // in increasing order.
+        let mut refit_node_indices: Vec<_> = {
+            let mut raw_indices: Vec<_> = refit_shape_indices.iter()
+                .map(|x| shapes[*x].bh_node_index())
+                .collect();
+
+            assert_eq!(raw_indices.len(),
```
alexd2580: Removed.

src/bvh/optimization.rs
```diff
@@ -0,0 +1,1039 @@
...
+            });
+
+            raw_indices.iter()
+                .map(|x| OptimizationIndex::Refit(*x))
+                .collect()
+        };
+
+        // As long as we have refit nodes left, take the list of refit nodes
+        // with the greatest depth (sweep nodes) and try to rotate them all.
+        while refit_node_indices.len() > 0 {
```
alexd2580: Done.

src/bvh/optimization.rs
```diff
@@ -0,0 +1,1039 @@
...
+            let mut sweep_node_indices = Vec::new();
+            let max_depth = {
+                let last_node_index = refit_node_indices.last().unwrap();
+                self.nodes[last_node_index.index()].depth()
+            };
+            while refit_node_indices.len() > 0 {
```
alexd2580: Done.

src/bvh/optimization.rs
```diff
@@ -0,0 +1,1039 @@
...
+                              node_index: usize,
+                              shapes: &[Shape])
+                              -> Option<OptimizationIndex> {
+        info!("   [{}]\t", node_index);
+
+        let node_clone = self.nodes[node_index].clone();
+        match node_clone {
+            BVHNode::Leaf { parent_index, shape_index, .. } => {
+                // The current node is a leaf.
+                info!("IAmA leaf node. Queueing parent ({}). {:?}.",
```
alexd2580: Done.

src/bvh/optimization.rs
```diff
@@ -0,0 +1,1039 @@
...
+        let left_children_nodes = get_children_node_data(&self.nodes[child_l_index]);
+        let right_children_nodes = get_children_node_data(&self.nodes[child_r_index]);
+
+        // Stores the Rotation that would result in the surface area best_surface_area,
+        // thus being the favored rotation that will be executed after considering all rotations.
+        // Consider all the possible rotations, choose the one with the lowest SAH cost.
+        let mut best_rotation: Option<(usize, usize)> = None;
+        {
+            let mut consider_rotation = |new_rotation: (usize, usize), sa: f32| {
+                info!("        SA {} vs. current SA {}.", sa, best_surface_area);
```
alexd2580: Done.

src/bvh/optimization.rs
```diff
@@ -0,0 +1,1039 @@
...
+                if sa < best_surface_area {
+                    best_surface_area = sa;
+                    best_rotation = Some(new_rotation);
+                }
+            };
+
+            // Child to grandchild rotations
+            if let Some((child_rl, child_rr)) = right_children_nodes {
+                {
+                    let sa = child_rl.aabb.surface_area() +
```
alexd2580: Each line is different, and it brings no merit to define two highly specialized functions, which would wrap this. Actually you can't wrap this in a more concise and selfexplanatory way.

src/bvh/optimization.rs
```diff
@@ -0,0 +1,1039 @@
...
+            *self.nodes[node_index].child_l_aabb_mut() = child_l_aabb;
+            *self.nodes[node_index].child_r_aabb_mut() = child_r_aabb;
+
+            /// Returns true randomly, with a chance given by the parameter
+            fn chance(chance: f32) -> bool {
+                let mut rng = thread_rng();
+                rng.next_f32() < chance
+            }
+
+            if node_index > 0 {
```
alexd2580: Done.

src/bvh/optimization.rs
```diff
@@ -0,0 +1,1039 @@
...
+            let mut node_count = 0usize;
+            self.assert_consistent_subtree(0, 0, &space, 0, &mut node_count);
+
+            // Check if all nodes have been counted from the root node.
+            // If this assert fails, it means we have a detached subtree.
+            assert_eq!(node_count, self.nodes.len(), "Detached subtree.");
+        }
+    }
+
+    impl PartialEq for BVHNode {
```
alexd2580: `useful`

alexd2580: I see no merit in defining this outside of the tests at all, but ok.

src/bvh/optimization.rs
```diff
@@ -0,0 +1,1039 @@
...
+        };
+
+        // Get indices and AABBs of all grandchildren
+        let left_children_nodes = get_children_node_data(&self.nodes[child_l_index]);
+        let right_children_nodes = get_children_node_data(&self.nodes[child_r_index]);
+
+        // Stores the Rotation that would result in the surface area best_surface_area,
+        // thus being the favored rotation that will be executed after considering all rotations.
+        // Consider all the possible rotations, choose the one with the lowest SAH cost.
+        let mut best_rotation: Option<(usize, usize)> = None;
```
alexd2580: This can transitively lead to a decrease in performance, because either the aabbs of the nodes, which are required in this function (which you suggest) would have to be cached in the nodes, or they would have to be recomputed veeery often. And computing the aabb of a complex shape can be quite difficult.

alexd2580: Or the interface to that function will be disgusting.

src/bvh/optimization.rs
Prior99: You are right.

src/bvh/optimization.rs
alexd2580: Done, review.

src/bvh/optimization.rs
alexd2580: Refactored, review, please.

src/bvh/optimization.rs
```diff
@@ -0,0 +1,1039 @@
...
+        let child_l_aabb = self.nodes[child_l_index].get_node_aabb(shapes);
+        let child_r_aabb = self.nodes[child_r_index].get_node_aabb(shapes);
+
+        // Contains the surface area that would result from applying the currently favored rotation.
+        // The rotation with the smallest SA will be applied in the end.
+        // The value is calculated by `child_l_aabb.surface_area() + child_r_aabb.surface_area()`.
+        let mut best_surface_area = child_l_aabb.surface_area() + child_r_aabb.surface_area();
+
+        #[derive(Debug, Copy, Clone)]
+        struct NodeData {
```
alexd2580: Done.

PR #27, opened 2019-01-14
Enable usage on stable rust

mjkoo:
Compiling on stable complains due to `#![feature(test)]`, this appears to be the only nightly feature used and its limited to the benchmark suite. Put this behind a "nightly" feature gate to enable usage on stable.

Perhaps "bench" or something similar is a more appropriate name for the feature given it's scope, and perhaps this should not be on by default (set to default now to match current behavior).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ language: rust
 rust:
   - stable
   - beta
-  - nightly
 matrix:
-  allow_failures:
-    - rust: stable
-    - rust: beta
+  include:
+    - rust: nightly
+      env: CARGO_TEST_FLAGS="--features bench"
+script: cargo test --verbose $CARGO_TEST_FLAGS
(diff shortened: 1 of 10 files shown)


svenstaro:
I think in that case it'd make sense to actually make this feature non-default and in the same go document how to run the benchmarks.

svenstaro:
In that case also remove the allowed failures from travis but run the benchmarks on travis while building nightly only.

mjkoo:
Cool, requested changes made

svenstaro:
Very cool, thanks!

svenstaro:
While you're at it you could also update it to rust 2018 if you're feeling up for it :P

mjkoo:
Sure I'll take a look :+1:

svenstaro:
Sweet!

PR #30, opened 2019-02-09
Add a function to retrieve the shape index from a leaf BVH node

h3r2tic:
Hey!

I'm using your crate to feed a BVH to the GPU with some custom packing/flattening. To do that, I needed to add a function to retrieve the shape index from a leaf node.

Cheers, and thanks for sharing your code!

### DIFF
--- a/src/bvh/bvh.rs
+++ b/src/bvh/bvh.rs
@@ -200,11 +200,12 @@ impl BVHNode {
         }
     }
 
-    /// Returns the index of the shape contained in the node, assuming it is a leaf.
-    pub fn shape_index(&self) -> usize {
+    /// Returns the index of the shape contained within the node if is a leaf,
+    /// or None if it is an interior node.
+    pub fn shape_index(&self) -> Option<usize> {
         match *self {
-            BVHNode::Leaf { shape_index, .. } => shape_index,
-            _ => panic!("Tried to get the shape index of an interior node."),
+            BVHNode::Leaf { shape_index, .. } => Some(shape_index),
+            _ => None,
         }
     }
 
@@ -709,7 +710,7 @@ impl BoundingHierarchy for BVH {
 
 #[cfg(test)]
 mod tests {
-    use crate::bvh::{BVHNode, BVH};
+    use crate::bvh::BVH;
     use crate::testbase::{build_some_bh, traverse_some_bh};
 
     #[test]
@@ -736,8 +737,8 @@ mod tests {
         let mut found_shapes = HashSet::new();
 
         for node in bh.nodes.iter() {
-            if let BVHNode::Leaf { .. } = node {
-                found_shapes.insert(node.shape_index());
+            if let Some(shape_index) = node.shape_index() {
+                found_shapes.insert(shape_index);
             }
         }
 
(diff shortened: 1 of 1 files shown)


svenstaro:
Can you make a test for that?

h3r2tic:
Yep, no problem.

src/bvh/bvh.rs
```diff
@@ -200,6 +200,14 @@ impl BVHNode {
         }
     }
 
+    /// Returns the index of the shape contained in the node, assuming it is a leaf.
+    pub fn shape_index(&self) -> usize {
+        match *self {
+            BVHNode::Leaf { shape_index, .. } => shape_index,
+            _ => panic!("Tried to get the shape index of an interior node."),
```
svenstaro: I'm not sure this should panic. Can't we rather define this to be `None` and return and `Option`?

h3r2tic: Yeah, that does sound better. One sec!

svenstaro: `panic!` should happen if you get run an invalid operation and I can't quite decide whether this is an invalid operation or whether it's actually valid to run this on non-`Leaf` nodes. What do you think?

svenstaro:
By the way, how are you using this exactly? :)

src/bvh/bvh.rs
h3r2tic: I think Option is a bit prettier. I wasn't too fond of having to deconstruct the enum just to test whether it's a node or a leaf, *and then* having the `shape_index` fn do another runtime check of self's variant.

svenstaro:
Also please make a test case for the `None` case. Should be easy. :)

h3r2tic:
Done!

As for how I'm using it, I'm packing the BVH into a format that only uses one child index (the other one follows directly in memory), and doing stackless traversal on the GPU. And then I'm generating five more of those trees -- reusing nodes, only flattening differently. They're specialized for the ray traveling more or less in +X, -X, +Y, -Y, +Z, -Z directions. The intersection shader picks the correct tree at the start.

It's based on the [ideas of Toshiya Hachisuka](https://www.ci.i.u-tokyo.ac.jp/~hachisuka/tdf2015.pdf).

Here's some very rough work in progress if you're curious: https://github.com/h3r2tic/rtoy-samples/blob/master/src/bin/raytrace/main.rs
Along with the corresponding shader: https://github.com/h3r2tic/rtoy-samples/blob/master/assets/shaders/raytrace.glsl

svenstaro:
Super sweet reference! I'm going to be using that for my next path tracer. I've already done a few in the past but didn't see that document yet.

src/bvh/bvh.rs
```diff
@@ -200,6 +200,15 @@ impl BVHNode {
         }
     }
 
+    /// Returns the index of the shape contained within the node if is a leaf,
+    /// or None if it is an interior node.
```
svenstaro: Tiny nitpick: `None` instead of None.

h3r2tic: :+1:  fixed.

h3r2tic:
Toshiya is a legend :) I'm a bit of an internet stalker of his.

svenstaro:
I'll pop out 0.3.1 with your changes once travis is happy. Feel free to submit more stuff to this. :)

h3r2tic:
Awesome, thanks! :)

PR #34, opened 2019-04-09
Make FlatNode fields public. Fix few doc links

### LINKED ISSUE #33: FlatNode's fields should be public.
Due to private fields and no access methods `FlatNode` cannot be used on GPU.

omni-viral:
Fix #33


svenstaro:
Thanks!

Issue #42, opened 2020-04-23
More general queries


medakk:
Would it be possible to add queries that are not tied to ray intersection tests? Something like can I provide a sphere and get a list of potential collisions to test for?

svenstaro:
That might be pretty cool but I don't see myself implementing this anytime soon. However, I'd accept a PR about it.

PR #43, opened 2020-04-24
[WIP] Add traversal queries based on points

medakk:
For issue #42 

This is just some quick code I whipped up to get it working for my use case. It includes a traversal for the points, and it would make sense to add traverse for any `Bounded` object. What would be the best way to generalize the traversal for all these different use cases?

One way would be to have a single traverse function that takes an argument `f`, which is a function that takes an AABB and returns whether or not there was a "hit". So, the ray traversal could be implemented as:

```rust
    pub fn traverse<'a, Shape: Bounded>(&'a self, ray: &Ray, shapes: &'a [Shape]) -> Vec<&Shape> {
        let mut indices = Vec::new();
        BVHNode::traverse_recursive(&self.nodes, 0, |&aabb| ray.intersects(aabb), &mut indices);
        indices
            .iter()
            .map(|index| &shapes[*index])
            .collect::<Vec<_>>()
    }
```

### DIFF
--- a/src/bounding_hierarchy.rs
+++ b/src/bounding_hierarchy.rs
@@ -2,6 +2,7 @@
 
 use crate::aabb::Bounded;
 use crate::ray::Ray;
+use nalgebra::Point3;
 
 /// Describes a shape as referenced by a [`BoundingHierarchy`] leaf node.
 /// Knows the index of the node in the [`BoundingHierarchy`] it is in.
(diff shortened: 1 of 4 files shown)


svenstaro:
@finnbear what do you think about this? It's pretty old at this point.

svenstaro:
Excellent, I'm staying tuned!

svenstaro:
Superseded by #108.

finnbear:
*Actually, superseded by #128, which allows you to query for AABB's that contain a point.

#108 adds a different type of query, that finds the closest shapes.

PR #61, opened 2021-04-28
Add traverse_iterator to BVH as more efficient alternative to traverse() -> Vec<Shape>

anjinl:
I noticed while profiling [my simple Rust implementation of "Ray tracing in one weekend"](https://github.com/anjinl/rust_one_weekend) a considerable CPU hotspot was heap alloc/free of memory around BVH.traverse(). This was because traverse() returns a Vec::new for every call.

I implemented an alternative approach to traverse a BVH. I added traverse_iterator() which returns Iterator instead of a Vec, so the calling function can iterate over hit shapes without any heap allocations.

The code is straightforward, but I'm very new to Rust so please critique it. I also couldn't figure out how to wrangle all the traits to make iterators part of BoundingHierarchy so it's only impl BVH for now.

Benchmark results on a Ryzen 3600x of intersecting sponza with 128 rays per iteration:
```
test bvh::iter::bench::bench_intersect_128rays_sponza_vec                ... bench:     174,705 ns/iter (+/- 5,857)
test bvh::iter::bench::bench_intersect_128rays_sponza_iter               ... bench:     150,897 ns/iter (+/- 6,722)
```

The perf gain is even larger on my hobby project, with perf going from ~15 million rays/sec to ~20 million rays/sec. It's a much simpler scene and aggressively multithreaded so perhaps allocation overhead hits harder.

### DIFF
--- a/src/bvh/bvh_impl.rs
+++ b/src/bvh/bvh_impl.rs
@@ -9,6 +9,7 @@ use crate::bounding_hierarchy::{BHShape, BoundingHierarchy};
 use crate::ray::Ray;
 use crate::utils::{concatenate_vectors, joint_aabb_of_shapes, Bucket};
 use crate::EPSILON;
+use crate::bvh::iter::BVHIterator;
 use nalgebra::Point3;
 use std::f32;
 use std::iter::repeat;
@@ -436,6 +437,17 @@ impl BVH {
             .iter()
             .map(|index| &shapes[*index])
             .collect::<Vec<_>>()
+    }
+
+    /// Creates a [`BVHIterator`] to traverse the [`BVH`].
+    /// Returns a subset of `shapes`, in which the [`AABB`]s of the elements were hit by `ray`.
+    ///
+    /// [`BVH`]: struct.BVH.html
+    /// [`AABB`]: ../aabb/struct.AABB.html
+    ///
+    pub fn traverse_iterator<'a, Shape: Bounded>(&'a self, ray: &'a Ray, shapes: &'a [Shape]) -> BVHIterator<Shape>
+    {
+        BVHIterator::new(self, ray, shapes)
     }
 
     /// Prints the [`BVH`] in a tree-like visualization.
--- a/src/bvh/bvh_impl.rs
+++ b/src/bvh/bvh_impl.rs
@@ -6,7 +6,7 @@
 
 use crate::aabb::{Bounded, AABB};
 use crate::bounding_hierarchy::{BHShape, BoundingHierarchy};
-use crate::bvh::iter::BVHIterator;
+use crate::bvh::iter::BVHTraverseIterator;
 use crate::ray::Ray;
 use crate::utils::{concatenate_vectors, joint_aabb_of_shapes, Bucket};
 use crate::EPSILON;
@@ -439,7 +439,7 @@ impl BVH {
             .collect::<Vec<_>>()
     }
 
-    /// Creates a [`BVHIterator`] to traverse the [`BVH`].
+    /// Creates a [`BVHTraverseIterator`] to traverse the [`BVH`].
     /// Returns a subset of `shapes`, in which the [`AABB`]s of the elements were hit by `ray`.
     ///
     /// [`BVH`]: struct.BVH.html
@@ -449,8 +449,8 @@ impl BVH {
         &'a self,
         ray: &'a Ray,
         shapes: &'a [Shape],
-    ) -> BVHIterator<Shape> {
-        BVHIterator::new(self, ray, shapes)
+    ) -> BVHTraverseIterator<Shape> {
+        BVHTraverseIterator::new(self, ray, shapes)
     }
 
     /// Prints the [`BVH`] in a tree-like visualization.
--- a/src/bvh/bvh_impl.rs
+++ b/src/bvh/bvh_impl.rs
@@ -6,10 +6,10 @@
 
 use crate::aabb::{Bounded, AABB};
 use crate::bounding_hierarchy::{BHShape, BoundingHierarchy};
+use crate::bvh::iter::BVHIterator;
 use crate::ray::Ray;
 use crate::utils::{concatenate_vectors, joint_aabb_of_shapes, Bucket};
 use crate::EPSILON;
-use crate::bvh::iter::BVHIterator;
 use nalgebra::Point3;
 use std::f32;
 use std::iter::repeat;
@@ -445,8 +445,11 @@ impl BVH {
     /// [`BVH`]: struct.BVH.html
     /// [`AABB`]: ../aabb/struct.AABB.html
     ///
-    pub fn traverse_iterator<'a, Shape: Bounded>(&'a self, ray: &'a Ray, shapes: &'a [Shape]) -> BVHIterator<Shape>
-    {
+    pub fn traverse_iterator<'a, Shape: Bounded>(
+        &'a self,
+        ray: &'a Ray,
+        shapes: &'a [Shape],
+    ) -> BVHIterator<Shape> {
         BVHIterator::new(self, ray, shapes)
     }
 
--- a/src/bvh/iter.rs
+++ b/src/bvh/iter.rs
@@ -1,5 +1,5 @@
 use crate::aabb::Bounded;
-use crate::bvh::{BVH, BVHNode};
+use crate::bvh::{BVHNode, BVH};
 use crate::ray::Ray;
 
 /// Iterator traverse a BVH without memory allocations
@@ -25,7 +25,7 @@ impl<'a, Shape: Bounded> BVHIterator<'a, Shape> {
 
             /// Reference to the input shapes array
             shapes: shapes,
-            
+
             /// Traversal stack. 4 billion items seems enough?
             stack: [0; 32],
 
@@ -141,9 +141,9 @@ mod tests {
     use crate::bvh::{BVHNode, BVH};
     use std::collections::HashSet;
     use std::f32;
-    
+
     use nalgebra::{Point3, Vector3};
-    
+
     use crate::aabb::{Bounded, AABB};
     use crate::bounding_hierarchy::BHShape;
     use crate::ray::Ray;
@@ -291,7 +291,6 @@ mod tests {
         }
     }
 
-
     #[test]
     /// Runs some primitive tests for intersections of a ray with a fixed scene given as a BVH.
     fn test_traverse_bvh() {
@@ -330,9 +329,7 @@ mod tests {
 #[cfg(all(feature = "bench", test))]
 mod bench {
     use crate::bvh::BVH;
-    use crate::testbase::{
-        load_sponza_scene, create_ray,
-    };
+    use crate::testbase::{create_ray, load_sponza_scene};
 
     #[bench]
     /// Benchmark the traversal of a `BVH` with the Sponza scene with Vec return.
@@ -342,13 +339,12 @@ mod bench {
 
         let mut seed = 0;
         b.iter(|| {
-            for _ in 0..128
-            {
+            for _ in 0..128 {
                 let ray = create_ray(&mut seed, &bounds);
-    
+
                 // Traverse the `BVH` recursively.
                 let hits = bvh.traverse(&ray, &triangles);
-        
+
                 // Traverse the resulting list of positive `AABB` tests
                 for triangle in &hits {
                     ray.intersects_triangle(&triangle.a, &triangle.b, &triangle.c);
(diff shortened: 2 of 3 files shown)


svenstaro:
Oh, this is great! The implementation looks very clean at a first glance. Let me have a deeper look.

anjinl:
Thanks. I just noticed I didn't run cargo fmt on the code, so I just did that on my branch.

src/bvh/bvh_impl.rs
```diff
@@ -438,6 +439,20 @@ impl BVH {
             .collect::<Vec<_>>()
     }
 
+    /// Creates a [`BVHIterator`] to traverse the [`BVH`].
+    /// Returns a subset of `shapes`, in which the [`AABB`]s of the elements were hit by `ray`.
+    ///
+    /// [`BVH`]: struct.BVH.html
+    /// [`AABB`]: ../aabb/struct.AABB.html
+    ///
+    pub fn traverse_iterator<'a, Shape: Bounded>(
```
svenstaro, line 448: Can you check whether anonymous lifetimes `'_` would be ok to use here? I don't think they were available back when this code was made. You don't have to change it here necessarily at this point but I'd just like to know whether it'd be an option as it'd make a great follow-up. :)

src/bvh/iter.rs
```diff
@@ -0,0 +1,377 @@
...
+            has_node: true,
+        }
+    }
+
+    /// Test if stack is empty.
+    fn is_stack_empty(&self) -> bool {
+        return self.stack_size == 0;
+    }
+
+    /// Push node onto stack. Not guarded against overflow.
```
svenstaro, line 48: Please document panic behavior (here and other overflowing methods) as per https://rust-lang.github.io/api-guidelines/documentation.html#function-docs-include-error-panic-and-safety-considerations-c-failure

I know this is not a public function but it will still help people reading this code.

At least I think an overflow will panic in Rust. To be honest, I'm not sure, it might just wrap. Something to investigate.

src/bvh/iter.rs
```diff
@@ -0,0 +1,377 @@
...
+        self.stack_size += 1;
+    }
+
+    /// Pop the stack and return the node. Not guarded against underflow.
+    fn stack_pop(&mut self) -> usize {
+        self.stack_size -= 1;
+        return self.stack[self.stack_size];
+    }
+
+    /// Attempt to move to the left child of the current node.
```
svenstaro, line 60: "Attempt" implies that it can fail but it seems it can't really fail or at least not that I can see. Please document what "attempt" means here and what an unsuccessful action looks like in this function and `move_right`.

src/bvh/iter.rs
```diff
@@ -0,0 +1,377 @@
...
+                // Go back up the stack and see if a node or leaf was pushed.
+                self.node_index = self.stack_pop();
+                match self.bvh.nodes[self.node_index] {
+                    BVHNode::Node { .. } => {
+                        // If a node was pushed, now attempt to move to its right child.
+                        self.move_right();
+                    }
+                    BVHNode::Leaf { shape_index, .. } => {
+                        // We previously pushed a leaf node. This is the "visit" of the in-order traverse.
+                        // Next time we call next we try to pop the stack again.
```
svenstaro, line 126: > // Next time we call `next()`, we try to pop the stack again.

src/bvh/iter.rs
```diff
@@ -0,0 +1,377 @@
...
+                        return Some(&self.shapes[shape_index]);
+                    }
+                }
+            }
+        }
+        return None;
+    }
+}
+
+// Copy of part of the BH testing in testbase.
```
svenstaro, line 137: Any chance you could unify this back to testbase? I think we're going to see less duplication that way. See, for instance, [here](https://github.com/svenstaro/bvh/blob/master/src/bvh/optimization.rs#L520) where we re-use a lot of stuff from testbase. This might be helpful.

svenstaro (changes requested):
This is looking great overall! I'd appreciate a bit less duplication in the test module. I think you can re-use a lot of existing stuff there and part of the benchmarks seem duplicated to me.

src/bvh/iter.rs
anjinl: Yes, I tried actually. The stumbling block for me was testbase functions all talk to BoundingHierarchy, but my traverse_iterator() is only on BVH. If you have a suggestion of how to express traverse_iterator on BoundingHierarchy I would absolutely implement it there (and for flat_bvh etc.) so we don't have this weird BVH-only traversal. But when I tried to add it on BoundingHierarchy, the rust compiler wanted me to suddenly throw "dyn"s everywhere. My lack of Rust knowledge let me down.

svenstaro: `dyn` is usually only associated with [trait objects](https://doc.rust-lang.org/book/ch17-02-trait-objects.html) so Rust somehow thought you were going for a trait object rather than an implementation. Maybe your syntax was slightly off?

anjinl:
Thanks for the suggestions by the way. I'll make those changes soon, and I'll attempt to make this "clippy" happy too :)

svenstaro:
> Thanks for the suggestions by the way. I'll make those changes soon, and I'll attempt to make this "clippy" happy too :)

Clippy is actually pretty cool. I hope you'll come around to appreciate its nagging. :P

src/bvh/iter.rs
anjinl: So underflowing stack_size will panic: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8f8f26d84ff38a9f4c111abb9f05911a

I think an "overflow" in this case (stack_size >= 32) will just trigger an OOB panic.

I'll add the comment on both functions.

src/bvh/iter.rs
anjinl: The obstacle I hit was: `traverse_iterator` on `BoundingHierarchy` must return _something_... It can't return `BVHIterator`. Like a base `TraverseIterator`... But, as soon as I return a trait, rust warns me:
warning: trait objects without an explicit `dyn` are deprecated

I guess that's just how it is though, without dynamic dispatch the runtime doesn't know what kind of iterator we're dealing with when looping over the results of `traverse_iterator` on a `BoundingHierarchy` 🤷‍♂️  .. Now I just need to figure out how to expose a generic iterator in `BoundingHierarchy` and the actual specialized iterators for `BVH`/`FlatBVH`

I like `TraverseIterator` and `BVHTraverseIterator` by the way. Adding "Traverse" feels more descriptive. Should I follow this naming?

svenstaro: Sure, go for it.

anjinl: Well, I went down the rabbit hole with `BoundingHierarchy` having a `traverse_iterator() -> Iterator<Shape: BHShape>` and I landed with a solution which depends on `#![feature(generic_associated_types)]` .. Hah. So the only way I could expose an iterator on `BoundingHierarchy` without lots of boxing and `dyn` is through a type in the `BoundingHierarchy`. Specifically this construction:

```
    type Iter<Shape: BHShape>: Iterator<Item = Shape>;
    fn traverse_iterator<'a, Shape: BHShape>(&'a self, ray: &Ray, shapes: &'a [Shape]) -> Self::Iter<Shape>;
```

This allows implementations like `BVH` to do:
```
    type Iter<Shape: BHShape> = BVHTraverseIterator<Item = Shape>;
    fn traverse_iterator<'a, Shape: BHShape>(&'a self, ray: &Ray, shapes: &'a [Shape]) -> Self::Iter<Shape>;
```

Unfortunately, it depends on `GAT` which you may not want to depend on yet...
```
error[E0658]: generic associated types are unstable
   --> src/bounding_hierarchy.rs:168:5
    |
168 |     type Iter<Shape: BHShape>: Iterator<Item = Shape>;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #44265 <https://github.com/rust-lang/rust/issues/44265> for more information
    = help: add `#![feature(generic_associated_types)]` to the crate attributes to enable
```

So I think I'd just keep traverse_iterator() on BVH only for now. Anyway I'll try to come up with some better way of sharing code with testbase :)

svenstaro: That was quite the adventure. I agree with your suggestion to keep the `traverse_iterator()` separate for the time being. We can always improve on this.

anjinl: Sounds good. I've just removed as much of the duplicated testbase code as I could. It's still a little verbose with its own `traverse_some_bvh` etc. But still much cleaner.

src/bvh/bvh_impl.rs
anjinl: I'm not 100% sure, but I don't think it applies to the new iterator code..

`impl<'a, Shape: Bounded> BVHTraverseIterator<'a, Shape> {`
needs a named lifetime parameter for
`pub fn new(bvh: &'a BVH, ray: &'a Ray, shapes: &'a [Shape]) -> Self`

And
`impl<'a, Shape: Bounded> Iterator for BVHTraverseIterator<'a, Shape> {`
needs it for
`type Item = &'a Shape;`

So I'm not sure if the anonymous syntax can be used anywhere. Unless I misunderstand the Rust 2018 docs.

svenstaro:
I just ran the code on my machine and I'm barely seeing a difference in performance:
```
test bvh::iter::bench::bench_intersect_128rays_sponza_iter               ... bench:     154,647 ns/iter (+/- 2,044)
test bvh::iter::bench::bench_intersect_128rays_sponza_vec                ... bench:     156,826 ns/iter (+/- 5,299)
```

Same numbers even with `RUSTFLAGS="-Ctarget-cpu=native" cargo bench --features bench`.

I've got a Ryzen Threadripper 3970X so our CPUs are very similar in architecture. I'm wondering why I'm not seeing a greater speedup.

anjinl:
That's very interesting! Are you on Linux perhaps? I wonder if the OS allocator is just way faster on Linux compared to my two platforms..? I've tested on 3 machines I use:
Windows 10 3600x
Windows 10 i7-5820K
MacBook Air Quad i5

Win10 3600x:
```
test bvh::iter::bench::bench_intersect_128rays_sponza_vec                ... bench:     193,437 ns/iter (+/- 2,947)
test bvh::iter::bench::bench_intersect_128rays_sponza_iter               ... bench:     154,602 ns/iter (+/- 6,441)
```

Win10 i7-5820K:
```
test bvh::iter::bench::bench_intersect_128rays_sponza_vec                ... bench:     260,601 ns/iter (+/- 203,922)
test bvh::iter::bench::bench_intersect_128rays_sponza_iter               ... bench:     234,075 ns/iter (+/- 203,418)
```

MacBook Air Quad-Core i5
```
test bvh::iter::bench::bench_intersect_128rays_sponza_vec                ... bench:     221,135 ns/iter (+/- 23,602)
test bvh::iter::bench::bench_intersect_128rays_sponza_iter               ... bench:     193,476 ns/iter (+/- 22,844)
```

If you are on Windows and you are not seeing the speedup, that would be interesting to root-cause... Also, bench here is single-threaded which is (depending on what's underlying heap alloc/free in Rust) the best-case for allocation overhead. I will add multithreaded benchmarks to run this comparison too.

svenstaro:
Yeah, I'm on Linux. I just kinda assumed you were, too. :)

`bench` is always single-threaded on purpose so that benchmarks do not influence each other by stealing resources or throttling the system by causing too much thermal stress so I'd not bother getting parallel benchmarks going.

svenstaro (approved):
Merging as is. I'd certainly appreciate follow-ups if you're still feeling up for it. The code is a little old in Rust terms and could certainly use some modern lovin'. :)

Also of course the duplication now introduced still isn't great. Perhaps we should just use proper integration tests now (they weren't available when this code was made). Baiscally files like `tests/test_hello.rs` will automatically be interpreted as separate integration test targets. This would clean up a ton!

svenstaro:
FYI I'd even consider making you a contributor if you'd like to work on this in a more involved capacity. :)

anjinl:
Great! I'm happy to contribute a bit more to both the iterator and `bvh`. My Rust knowledge is still quite limited, but this seems like a good way to grow it. I'm particularly interested in anything perf-related (if that wasn't obvious).

Regarding `bench`, I understand it, but I'd still be curious if you could try a traverse_vec/traverse_iter in parallel on your machine at some point. Code like this can of course behave very differently in single or multi threaded environment. So I wonder if you can see a bigger perf difference in Linux if there's more contention around the heap. If Linux can pull this off without a bigger delta I'd be really impressed 😄

svenstaro:
You might wanna give Linux a try then and see how things perform for you. :)

PR #63, opened 2021-04-30
Port from nalgebra to glam

### LINKED ISSUE #38: Support glam instead of or in addition to nalgebra?
glam (https://github.com/bitshifter/glam-rs) is a much slimmer, faster-to-build and neater vector math library than nalgebra, which is so complex that it often triggers Rust compiler bugs (a discussion to be had elsewhere).

It might be possible to support both through feature flags and some type redefinitions.

mdesmedt:
Ported the use of math types to glam. Mostly very straightforward. It was 99% search & replace. Tests and benchmarks pass.

I've opted to type alias `glam::Vec3` directly in lib.rs for visibility and simplicity. So now we have `bvh::Point3` and `bvh::Vector3` which are used everywhere.

Changes of interest:
`lib.rs` for the above type aliases
`aabb.relative_eq` (no longer using approx, because it's not implemented on Vec3)
`ray.sign` (split out the Vector3<usize> into 3 variables)

Resolves #38

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,10 +15,10 @@ license = "MIT"
 
 [dependencies]
 approx = "0.4"
-nalgebra = "0.26"
 rand = "0.8"
 log = "0.4"
 num = "0.4"
+glam = "0.13.0"
 
 [dev-dependencies]
 quickcheck = "0.9"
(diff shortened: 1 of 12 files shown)


mdesmedt:
Hmm. Note: Besides the straightforward port, we should investigate whether to use the aligned `Vec3A` instead of `Vec3` for potentially better SIMD code generation. I'll take a look.

svenstaro (approved):
I can roughly reproduce your results. This seems like a good overall compromise.

Issue #71, opened 2021-06-08
improvements


dbenson24:
Thanks for maintaining this crate, I've been integrating it into a game I'm working on in Unity and in doing so I've added/am adding a bunch of things that I needed and was wondering how much of it you might be interested me trying to separate out for a PR.
The performance is amazing!

I've been implementing some new query shapes
- Sphere
- Capsule
- AABB
- OBB (almost)
- Segment (tbd)

In addition to those I've also been adding some new methods
- Add (add a new node to the bvh without a rebuild)
- Remove (remove a node without a rebuild)
- Rebuild (rebuild bvh but reuse the vector to reduce allocations)

I've also defined a C FFI so that it could be used outside of Rust code

My version is also 64bit because I needed the precision

If any of this stuff sounds like something you might be interested in a PR for let me know.

svenstaro:
I'd certainly appreciate contributions! However, please try to split it up to one PR per shape if you can as that makes reviewing easier.

StarArawn:
I would be interested in the Add/Remove/Rebuild. Do you have a fork up somewhere I can take a look at?

dbenson24:
I will try and get it uploaded in the next day or two. I've pretty heavily edited a few things, a merge back shouldn't be too hard but I haven't prioritized the time for that yet

dbenson24:
also I haven't implemented rebuild yet, I know how I want to but my priority is has been on add and remove. Do note that heavy use of add and remove will cause a hefty toll on the locality of the bvh over time. Regardless I'm super happy with the performance of this BVH, currently I'm able to do ~100k raycasts against a bvh with a 100k items in it < 1ms or so

dbenson24:
https://github.com/dbenson24/bvh @StarArawn warning that I have done literally nothing to prep it for viewing other than migrating my changes back into a fork of the repo

dbenson24:
@svenstaro

So I've continued working on/improving my branch of your BVH as I'm using it for a unity game I'm working on. I've attached a bench run of the 32bit version below. At this point the code has diverged quite a lot from your current version. If you are interested I'm willing to work with you to upstream changes, otherwise I'm considering publishing a fork with the 32/64bit versions + the C FFI bindings.
General summary of changes is something like this:
Optimize the build function to reduce allocations
Merge the centroid calculation with the bucket building process
Parallelize build using Rayon
Use a SmallVec to back the iterator so that a panic during traversal is impossible
New shapes to query the BVH with as listed above
[...]

PR #80, opened 2022-01-14
optimizations and other improvements 

dbenson24:
#71 

Opening the PR, it's definitely not ready to be merged yet there's a couple more things I'm finishing up but I wanted to open earlier so if you wanted to read things over you'd be able to


bvh-lib/Cargo.toml
```diff
@@ -17,4 +17,14 @@ glam = "0.20"
...
 [lib]
 name="bvh_lib"
-crate-type = ["rlib", "dylib"]
\ No newline at end of file
+crate-type = ["rlib", "dylib"]
+
+[profile.release]
+lto = true
+codegen-units = 1
+debug = true
```
svenstaro, line 25: Do we want debug here?

bvh-lib/Cargo.toml
```diff
@@ -17,4 +17,14 @@ glam = "0.20"
...
+
+[profile.bench]
+lto = true
+codegen-units = 1
+debug = true
```
svenstaro, line 30: Do we want debug here?

svenstaro (changes requested):

svenstaro:
Is there any chance you could cut down the PR into smaller chunks? I realize you did a lot of refactoring but are there perhaps any smaller chunks to make reviewing easier?

dbenson24:
The FFI stuff can very easily be separated so I'll do that. I'm not even sure what the best way to distribute that is

I think I can also pull the new shapes out into a separate change, I'll see what I can come up with

bvh-f64/Cargo.toml
```diff
@@ -0,0 +1,59 @@
+[package]
+name = "bvh-f64"
```
svenstaro, line 2: Do we really need a specialized crate for this? Can't we use traits for this to have Rust do the specialization?

svenstaro (changes requested):

flamegraph.svg
```diff
@@ -0,0 +1,412 @@
+<?xml version="1.0" standalone="no"?><!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd"><svg version="1.1" width="1200" height="182" onload="init(evt)" viewBox="0 0 1200 182" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:fg="http://github.com/jonhoo/inferno"><!--Flame graph stack visualization. See https://github.com/brendangregg/FlameGraph for latest version, and http://www.brendangregg.com/flamegraphs.html for examples.--><!--NOTES: --><defs><linearGradient id="background" y1="0" y2="1" x1="0" x2="0"><stop stop-color="#eeeeee" offset="5%"/><stop stop-color="#eeeeb0" offset="95%"/></linearGradient></defs><style type="text/css">
```
svenstaro, line 1: Did you mean to commit this file? A Makefile target might be more appropriate to generate this or perhaps even just a README entry.

svenstaro (changes requested):

flamegraph.svg
dbenson24: Yea I thought I had already deleted this

bvh-f64/Cargo.toml
dbenson24: So I tried a few approaches to this. It might be a possibility with nalgebra backing things to have the float be a generic on everything, but with glam there's no trait to abstract dvec3 from vec3 so we would have to wrap that ourselves. This trick using separate crates in the workspace is something I got from the author of nalgebra and how he builds rapier (2d 3d 32 and 64bit). When the workspace is all setup with this cargo test runs the test for all the versions and you'll get compiler errors from rust analyzer for both at the same time

dbenson24:
I've been a bit busy last couple of weeks haven't had time to look at breaking this apart. if that's still something you would prefer I can start working on trying to do so

svenstaro:
Well if we can somehow do this in small incremental PRs that would be ideal but I can understand if you don't want to go down that path but it would make reviewing faster for me for sure. You get to decide. :)

dbenson24:
I've pulled the ffi bindings out, I can add them back in if you want them in the repo otherwise I will just throw them up in a separate repo.

In terms of reviewing this, the big change that affects the whole crate is refactoring all the types to use the vec/float types defined at the crate root. Otherwise the changes are pretty much file by file but because of that change to all of the types it's kind of a pain to break this down into little commits.

src/shapes/capsule.rs
```diff
@@ -0,0 +1,80 @@
...
+            len,
+        }
+    }
+}
+
+impl IntersectionAABB for Capsule {
+    fn intersects_aabb(&self, aabb: &AABB) -> bool {
+        /*
+        // Use Distance from closest point
+        let mut point: Vector3 = self.start;
```
Elabajaba, line 41: Commented out code. Delete it if it isn't used?

src/shapes/ray.rs
```diff
@@ -299,29 +331,97 @@ impl Ray {
...
+
+        let z_min = (aabb[self.sign_z].z - self.origin.z) * self.inv_direction.z;
+        let z_max = (aabb[1 - self.sign_z].z - self.origin.z) * self.inv_direction.z;
+
+        if (ray_min > z_max) || (z_min > ray_max) {
+            return None;
+        }
+
+        // Only required for bounded intersection intervals.
+        // if z_min > ray_min {
```
Elabajaba, line 384: More commented out code.

src/shapes/ray.rs
```diff
@@ -44,56 +45,54 @@ pub struct Ray {
...
-        Ray {
-            origin,
-            direction,
-            inv_direction: Vector3::new(1.0 / direction.x, 1.0 / direction.y, 1.0 / direction.z),
-            sign_x: (direction.x < 0.0) as usize,
-            sign_y: (direction.y < 0.0) as usize,
-            sign_z: (direction.z < 0.0) as usize,
-        }
+/// This trait can be implemented on anything that can intersect with a `Ray`
+pub trait IntersectionRay {
```
Elabajaba, line 81: Do you have your own IntersectionRay impls outside of this library? I'm wondering what the performance tradeoff of using a vtable here vs an enum would be.

Also, is there a reason why IntersectionRay is only implemented for sphere+triangle, but not aabb/capsule/obb?

bvh-f64/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+itertools = "0.10.1"
+serde = { version = "1", features = ["derive"] }
+glam = { version = "0.20", features = ["serde"] }
+serde_json = "1"
+
+[features]
+default = ["f64"]
+bench = []
+f64 = []
+# Unfortunately can't use "serde" as the feature name until https://github.com/rust-lang/cargo/issues/5565 lands
```
Elabajaba, line 48: This was stabilized in Rust 1.60.

bvh/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+criterion = "0.3"
+itertools = "0.10.1"
+serde = { version = "1", features = ["derive"] }
+glam = { version = "0.20", features = ["serde"] }
+serde_json = "1"
+
+[features]
+default = []
+bench = []
+# Unfortunately can't use "serde" as the feature name until https://github.com/rust-lang/cargo/issues/5565 lands
```
Elabajaba, line 46: This was stabilized in Rust 1.60.

src/mesh/convex_hull.rs
```diff
@@ -0,0 +1,5 @@
+use crate::Vector3;
```
Elabajaba, line 1: Empty file.

src/mesh/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod convex_hull;
```
Elabajaba, line 1: Also empty.

Elabajaba:
So I decided to try building a toy pathtracer against this branch, and I'm curious about how users are supposed to use the included shapes? 

Currently none of them can actually be used in the BVH since none of them have a BHShape impl (previously you only needed a Bounded impl, in this branch you need both a Bounded and a BHShape impl), so you can't actually include them in the BVH. If you're supposed to shoot rays and intersect them against the included shapes, it'd be nice if they used a Structure of Arrays (SoA) layout (a simple SoA layout gets me ~2x the performance with 100 spheres, or ~3x with 1000 spheres in my pathtracer), however because intersects_ray (from the IntersectionRay impl) doesn't have a way to return an index, you can't currently use your own SOA layouts.

I believe IntersectionAABB also doesn't have any way to return an index, so it also can't be used with SOA layouts.

src/testbase.rs
```diff
@@ -166,7 +163,8 @@ pub fn traverse_some_bh<BH: BoundingHierarchy>() {
 }
 
 /// A triangle struct. Instance of a more complex `Bounded` primitive.
-#[derive(Debug)]
+#[derive(Copy, Clone, Debug)]
+#[cfg_attr(feature = "serde_impls", derive(serde::Serialize, serde::Deserialize))]
 pub struct Triangle {
```
Elabajaba, line 168: I feel like this should be removed since this pr includes a triangle shape, and having 2 public triangle structs is confusing.

dbenson24:
@Elabajaba the shapes missing impls are just because I haven't had time to write that code. This PR stems entirely from me writing other code and adding stuff to BVH as I needed it. In terms of the shapes not being able to be used in SOA, I chose not to attach an index to any of the shapes because it's very easy if you want to do that yourself. Just define an IndexedShape<T> that stores the shape and it's index. I didn't think that it made sense to force an index onto every shape.

bvh-f64/Cargo.toml
marstaik: I would very much like to see this move to nalgebra, it would make the usage much cleaner via `Bvh<T>`. This would definitely be possible with the nalgebra vector types and number bounds.

Heck, because of the AABB's, assuming there are no dot products/cross products (which I don't think there are - haven't checked), it should work with integer types as well, and maybe even unsigned types (but there would have to be better bounds checking).

I could try and help get this working with nalgebra as well.

svenstaro: I think that would be pretty cool.

marstaik: I'm not sure how to approach this for now though, should I wait for these changes to get merged in?

marstaik: Hmmm it might be easier for me to port it now, and have this PR rebase on it. Otherwise we would be splitting code and then remerging.

Elabajaba: Has nalgebra fixed the issues that made this crate switch from nalgebra to glam previously?

svenstaro:
Because of the discussion in the other thread: I wouldn't mind a co-maintainer on this. I don't use it much myself currently and I think someone with more stakes in this could really drive it forward a bit.

marstaik:
> Because of the discussion in the other thread: I wouldn't mind a co-maintainer on this. I don't use it much myself currently and I think someone with more stakes in this could really drive it forward a bit.

I may be interested, I plan to use it quite a bit in the future. I am a bit new to the rust side of things, but I generally worked with c++ metaprogramming before my switch to game engine development in rust.

svenstaro:
> > Because of the discussion in the other thread: I wouldn't mind a co-maintainer on this. I don't use it much myself currently and I think someone with more stakes in this could really drive it forward a bit.
> 
> I may be interested, I plan to use it quite a bit in the future. I am a bit new to the rust side of things, but I generally worked with c++ metaprogramming before my switch to game engine development in rust.

How about you start off by making one or two PRs that I'll review and then I can make you contrib from there if we click well?

svenstaro:
@marstaik how do you wanna play this PR? Shall we close it? Shall we leave it open until you have implemented the bits that you are interested in? I don't think the original author is going to return to this any time soon and the rebase is very gnarly at this point.

marstaik:
I'm going to close this, the original author has already started porting the changes incrementally starting with this:
https://github.com/svenstaro/bvh/pull/99

PR #83, opened 2022-06-02
Fix readme example

Elabajaba:
The readme example didn't work. This adds a BHShape impl to the sphere so that it'll work. It also removes the BoundingHierarchy import from the lib.rs docs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -95,3 +95,6 @@ mod utils;
 
 #[cfg(test)]
 mod testbase;
+
+#[cfg(doctest)]
+doc_comment::doctest!("../README.md", readme);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,7 +15,7 @@
 //!
 //! ```
 //! use bvh::aabb::{AABB, Bounded};
-//! use bvh::bounding_hierarchy::{BoundingHierarchy, BHShape};
+//! use bvh::bounding_hierarchy::BHShape;
 //! use bvh::bvh::BVH;
 //! use bvh::{Point3, Vector3};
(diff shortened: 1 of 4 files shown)


svenstaro (commented):
Thanks! Perhaps this would be a good time to start using doccomment so that the readme will actually be compiled during CI as well. It should be very simple to add. Would you like to add that as well?

Elabajaba:
I setup doc-comment, but it's a bit ugly. Since it's essentially just doing an include_str!(README.md) in lib.rs, the line numbers for errors don't line up with where the lines are in README.md (and if you put it in /tests/readme.rs like some crates have, then it never actually gets run).

It also complains about all the benchmark results in the readme unless you remove the rust tag on their code blocks and add a different language (not choosing a language doesn't work either), so if you have a preferred colour scheme for the benchmarks let me know.

src/lib.rs
```diff
@@ -95,3 +95,9 @@ mod utils;
 
 #[cfg(test)]
 mod testbase;
+
+#[cfg(doctest)]
+use doc_comment::doctest;
+
+#[cfg(doctest)]
+doctest!("../README.md", readme);
```
svenstaro, line 103: Let's do it like this:

```suggestion
#[cfg(doctest)]
doc_comment::doctest!("../README.md", readme);
```

svenstaro (changes requested):
Thanks! Just a little thing.

src/lib.rs
Elabajaba: Done and squashed it into the doc-comment commit.

svenstaro:
Thanks, this is good stuff!

PR #87, opened 2022-09-19
Improve ray/AABB benchmarks

tavianator:
- Fix ray/AABB benchmarks
- Optimize ray/AABB intersection tests.

### DIFF
--- a/src/ray.rs
+++ b/src/ray.rs
@@ -482,64 +482,71 @@ mod bench {
     use rand::rngs::StdRng;
     use rand::{Rng, SeedableRng};
 
+    use test::{black_box, Bencher};
+
     use crate::aabb::AABB;
     use crate::ray::Ray;
 
     use crate::testbase::{tuple_to_point, tuple_to_vector, TupleVec};
 
-    /// Generates some random deterministic `Ray`/`AABB` pairs.
-    fn gen_random_ray_aabb(rng: &mut StdRng) -> (Ray, AABB) {
+    /// Generate a random deterministic `Ray`.
+    fn random_ray(rng: &mut StdRng) -> Ray {
+        let a = tuple_to_point(&rng.gen::<TupleVec>());
+        let b = tuple_to_vector(&rng.gen::<TupleVec>());
+        Ray::new(a, b)
+    }
+
+    /// Generate a random deterministic `AABB`.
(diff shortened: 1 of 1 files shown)


src/ray.rs
```diff
@@ -63,6 +63,16 @@ impl Intersection {
...
 
+/// Fast minimum that matches SSE NaN semantics.
+fn min(x: f32, y: f32) -> f32 {
+    if x < y { x } else { y }
+}
+
+/// Fast maximum that matches SSE NaN semantics.
+fn max(x: f32, y: f32) -> f32 {
+    if x > y { x } else { y }
+}
```
svenstaro, line 74: Perhaps force `#[inline]`?

svenstaro: It might also be worth retaining the original note about performance.

svenstaro:
Very good stuff, thanks! I always wondered about the benchmark results but never quite had the will to figure it out. Also great optimization. Do you reckon the optimization is only due to the compiler generating SSE for it now?

svenstaro:
Looks like you're going to have to appease the CI, though. :) After you do, I'm happy to merge as is and cut a release.

tavianator:
> Do you reckon the optimization is only due to the compiler generating SSE for it now?

The generated code for `x.min(y)` on baseline x86-64 is something like

```asm
        cmpunordss      xmm2, xmm0
        movaps  xmm3, xmm2
        andps   xmm3, xmm1
        minss   xmm1, xmm0
        andnps  xmm2, xmm1
        orps    xmm2, xmm3
```

For `if x < y { x } else { y }` it's just

```asm
        minss   xmm0, xmm1
```

svenstaro:
It's a bit silly that the built in is so inefficient but there's probably a good reason for it.

tavianator:
> It's a bit silly that the built in is so inefficient but there's probably a good reason for it.

Well, [these semantics](https://github.com/svenstaro/bvh/pull/87/files#diff-a3a9b4c318707d2160a0d62c757a060d873e6da159d7ba1653f3972bb4e1a88dR73-R76) are pretty surprising, so I get going with the more consistent ones for the stdlib

svenstaro:
@tavianator Could you rebase? I fixed the lints and the CI.

svenstaro:
Super thanks!

PR #93, opened 2022-10-30
Add blanket impls for `Bounded`, `BHShape`, `BoundingHierarchy`

nerosnm:
These implementations are useful to have for most traits generally, but in particular I was running into issues trying to call methods from `Bounded` where I had a `Vec<&T>` where `T: Bounded` (for example).

I've included blanket impls of:

- `Bounded` for `&T`, `&mut T` and `Box<T>`
- `BHShape` for `&mut T` and `Box<T>` (not `&T` because `set_bh_node_index()` takes `&mut self`)
- `BoundingHierarchy` for `Box<T>` (not `&T` or `&mut T` because `build()` is a constructor/needs to return an owned value)

### DIFF
--- a/src/aabb.rs
+++ b/src/aabb.rs
@@ -60,6 +60,24 @@ pub trait Bounded {
     fn aabb(&self) -> AABB;
 }
 
+impl<T: Bounded> Bounded for &T {
+    fn aabb(&self) -> AABB {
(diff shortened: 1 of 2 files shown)


svenstaro:
Sure, why not. Thanks!

PR #95, opened 2023-02-26
Widen lifetime bounds on BVHTraverseIterator

Steelbirdy:
This fixes an issue I was having where an enclosing function couldn't return data referencing the shape returned by the BVHTraverseIterator.


svenstaro:
Sadly CI breaks on tarpaulin at the moment. I'll have to fix that one of these days.

svenstaro:
Merged, thanks!

Issue #56, opened 2021-01-14
Support for Distance Queries


SallySoul:
This is sort of a feature request, but I would also be happy to work on implementing it!

I am experimenting with porting some of our C++ code to rust. I would love to use a community supported BVH implementation this time around. However, besides ray intersections, we also make extensive use of distance queries, i.e. what is the closest object/s in the BVH to arbitrary points.

Is this functionality y'all would be open to having in this crate?

I have some ideas on how that would look which I'd be happy to discuss, and there are some other queries I'd be willing to port over too like a Fast Winding Number, but I don't want to get ahead of myself or be overwhelming.

svenstaro:
I think having this kind of feature would be excellent and I'd happily accept a PR.

PR #98, opened 2023-05-03
Implement ordered traverse

dashedman:
Hi. I have implemented a new iterator to traverse.
It traverses the tree from the nearest AABB to the farthest for the passed ray.

PR #108, opened 2024-06-27
Add nearest shape to point queries for bvh

Azkellas:
I recently implemented a distance point query with a bvh for my crate mesh_to_sdf by trait extending this crate (see https://github.com/Azkellas/mesh_to_sdf/pull/75). This PR is an attempt to upstream the changes and would solve https://github.com/svenstaro/bvh/issues/56.

This PR is different from https://github.com/svenstaro/bvh/pull/43 which only returns shapes with aabb containing the query point. On the contrary, this PR returns a conservative list of candidates, one of which for sure contains the shape that is closest to the query point. This is needed since we can't know for sure if the query point is in the aabb of its closest object.

To do so, we keep track of the best minimum/maximum distance found yet in the tree. For each node we compute the aabb min/max distance from the point (the minimum is 0 if the point is inside). Then we can prune or investigate the subtrees depending on the aabb's min/max compared to the current best min/max.

This means the search visits a subtree and not a single linear branch. It might be interesting to add an `approx` version following https://github.com/svenstaro/bvh/pull/43 implementation which would be faster and "close enough".

I couldn't implement the `nearest_candidate` method for `FlatBvh`.  Pre-order no-stack no-recursion tree traversal requires to store the parent index in the node afaik. I didn't want to change the memory layout without talking about it first. This PR is in draft state for this reason.

Testing is a bit rudimentary (and no benchmark) as I did most of it in my crate and haven't ported it yet.
Anyway for 100k triangles and 15k queries (in mesh's bounding box) it is 10x faster than brute forcing while being exact.

Waht remains to be done:
- [x] Upstream tests and benches
- [x] Implement `nearest_candidate` for `FlatBvh`
- [x] Implement an `approx` version? At least test the performance gain.

Thank you for the work, this crate is very useful.

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -10,6 +10,7 @@ cargo-fuzz = true
 
 [dependencies]
 arbitrary = { version = "1.4.1", features = ["derive"] }
+approx = "0.5"
 libfuzzer-sys = "0.4"
 nalgebra = "0.33"
 ordered-float = { version = "4.6.0", features = ["arbitrary"] }
--- a/fuzz/fuzz_targets/fuzz.rs
+++ b/fuzz/fuzz_targets/fuzz.rs
@@ -27,6 +27,7 @@ use bvh::ball::Ball;
 use bvh::bounding_hierarchy::{BHShape, BoundingHierarchy};
 use bvh::bvh::Bvh;
 use bvh::flat_bvh::FlatBvh;
+use bvh::point_query::PointDistance;
 use bvh::ray::Ray;
 use libfuzzer_sys::fuzz_target;
 use nalgebra::{Point, SimdPartialOrd};
@@ -143,6 +144,14 @@ impl<const D: usize> BHShape<Float, D> for ArbitraryShape<D> {
 impl<const D: usize> ArbitraryShape<D> {
     fn mode_is_grid(&self) -> bool {
         self.a.mode.is_grid() && self.b.mode.is_grid()
+    }
+}
+
+impl<const D: usize> PointDistance<Float, D> for ArbitraryShape<D> {
+    fn distance_squared(&self, query_point: Point<Float, D>) -> Float {
+        let center = self.aabb().center();
+        let offset = query_point - center;
+        offset.dot(&offset)
     }
 }
 
@@ -270,7 +279,7 @@ struct Workload<const D: usize> {
     shapes: Vec<ArbitraryShape<D>>,
     /// Traverse by ray.
     ray: ArbitraryRay<D>,
-    /// Traverse by point.
+    /// Traverse by point and for nearest_to query.
     point: ArbitraryPoint<D>,
     /// Traverse by AABB.
     aabb: ArbitraryShape<D>,
@@ -317,6 +326,51 @@ impl<const D: usize> Workload<D> {
         }
 
         traverse
+    }
+
+    /// Compares normal and flat-BVH nearest_to query.
+    ///
+    /// # Panics
+    /// Panics if the results differ in an unexpected way.
+    fn fuzz_nearest_to<'a>(
+        &'a self,
+        bvh: &'a Bvh<Float, D>,
+        flat_bvh: &'a FlatBvh<Float, D>,
+        query_point: Point<Float, D>,
+    ) {
+        let best = bvh.nearest_to(query_point, &self.shapes);
+        let flat_best = flat_bvh.nearest_to(query_point, &self.shapes);
+
+        // Assert we get None if and only if the bvh is empty.
+        assert_eq!(best.is_none(), self.shapes.is_empty());
+        assert_eq!(flat_best.is_none(), self.shapes.is_empty());
+
+        // Asserting equality between the two results return many false positives
+        // where two differents shapes give the same distance due to numerical approximations
+        // e.g.
+        // left: Some((ByPtr((Aabb { min: [-5000.0, 0.0, 0.0], max: [5000.0, 1.0, 1.0] }, [5000.0, 0.0, 0.0])), 5000.0))
+        // right: Some((ByPtr((Aabb { min: [5.418294e-39, -5000.0, 9.1834e-41], max: [5.418294e-39, -5000.0, 9.1834e-41] }, [-5000.0, 9.1834e-41, 0.0])), 5000.0))
+        // assert_eq!(best, flat_best);
+        // as such we only assert for distance equivalence if they are not None
+        if let Some((best, flat_best)) = best.zip(flat_best) {
+            approx::assert_abs_diff_eq!(best.1, flat_best.1);
+
+            // Brute force search for the ground truth, shapes is not empty since the results are not None.
+            let mut bruteforce = (
+                &self.shapes[0],
+                self.shapes[0].distance_squared(query_point),
+            );
+            for shape in &self.shapes[1..] {
+                let distance = shape.distance_squared(query_point);
+                if distance < bruteforce.1 {
+                    bruteforce = (shape, distance);
+                }
+            }
+            bruteforce = (bruteforce.0, bruteforce.1.sqrt());
+            // Again we only test for distances.
+            approx::assert_abs_diff_eq!(best.1, bruteforce.1);
+            approx::assert_abs_diff_eq!(flat_best.1, bruteforce.1);
+        }
     }
 
     /// Called directly from the `cargo fuzz` entry point. Code in this function is
@@ -474,6 +528,9 @@ impl<const D: usize> Workload<D> {
                 nearest_child_traverse_iterator,
                 farthest_child_traverse_iterator
             );
+
+            // Check that these don't panic and return the same value.
+            self.fuzz_nearest_to(&bvh, &flat_bvh, self.point.point());
 
             if let Some(mutation) = self.mutations.pop() {
                 match mutation {
--- a/src/aabb/aabb_impl.rs
+++ b/src/aabb/aabb_impl.rs
@@ -594,6 +594,53 @@ impl<T: BHValue, const D: usize> Aabb<T, D> {
     pub fn largest_axis(&self) -> usize {
         self.size().imax()
     }
+
+    /// Returns the minimum and maximum distances to the [`Aabb`].
+    /// The minimum distance is the distance to the closest point on the box,
+    /// 0 if the point is inside the box.
+    /// The maximum distance is the distance to the furthest point in the box.
+    ///
+    /// # Examples
+    /// ```
+    /// use bvh::aabb::Aabb;
+    /// use nalgebra::Point3;
+    ///
+    /// let min = Point3::new(0.0,0.0,0.0);
+    /// let max = Point3::new(10.0,10.0,10.0);
+    ///
+    /// let aabb = Aabb::with_bounds(min, max);
+    /// let query = Point3::new(20.0, 0.0, 0.0);
+    /// let (min_dist, max_dist) = aabb.get_min_max_distances(&query);
+    /// assert!(min_dist == 10.0);
+    /// assert!(max_dist == (10.0_f32*10.0 + 10.0*10.0 + 20.0*20.0).sqrt());
+    /// ```
+    ///
+    /// [`Aabb`]: struct.Aabb.html
+    ///
+    pub fn get_min_max_distances(&self, point: &Point<T, D>) -> (T, T) {
+        let half_size = self.size() * T::from_f32(0.5).unwrap();
+        let center = self.center();
+
+        let delta = point - center;
+
+        // See <https://iquilezles.org/articles/distfunctions/>
+        let q = delta.abs() - half_size;
+        let outside_vec = q.map(|x| x.max(T::zero()));
+        let min_dist = outside_vec.dot(&outside_vec).sqrt(); // norm without requiring ComplexField.
+
+        // The signum helps to determine the furthest point
+        let signum = delta
+            // Invert the signum to get the furthest vertex
+            .map(|x| -x.signum())
+            // Make sure we're always on a vertex and not on a face if the point is aligned with the box
+            .map(|x| if x != T::zero() { x } else { T::one() });
+
+        let furthest = center + signum.component_mul(&half_size);
+        let furthest_delta = point - furthest;
+        let max_dist = furthest_delta.dot(&furthest_delta).sqrt();
+
+        (min_dist, max_dist)
+    }
 }
 
 /// Default instance for [`Aabb`]s. Returns an [`Aabb`] which is [`empty()`].
--- a/src/aabb/aabb_impl.rs
+++ b/src/aabb/aabb_impl.rs
@@ -619,7 +619,7 @@ impl<T: BHValue, const D: usize> Aabb<T, D> {
     ///
     pub fn get_min_max_distances(&self, point: &Point<T, D>) -> (T, T) {
         let half_size = self.size() * T::from_f32(0.5).unwrap();
-        let center = self.center();
+        let center = self.min + half_size;
 
         let delta = point - center;
 
@@ -631,9 +631,8 @@ impl<T: BHValue, const D: usize> Aabb<T, D> {
         // The signum helps to determine the furthest point
         let signum = delta
             // Invert the signum to get the furthest vertex
-            .map(|x| -x.signum())
-            // Make sure we're always on a vertex and not on a face if the point is aligned with the box
-            .map(|x| if x != T::zero() { x } else { T::one() });
+            // signum never returns 0.0 so we're sure to be on a vertex.
+            .map(|x| -x.signum());
 
         let furthest = center + signum.component_mul(&half_size);
         let furthest_delta = point - furthest;
--- a/src/aabb/aabb_impl.rs
+++ b/src/aabb/aabb_impl.rs
@@ -595,10 +595,9 @@ impl<T: BHValue, const D: usize> Aabb<T, D> {
         self.size().imax()
     }
 
-    /// Returns the minimum and maximum distances to the [`Aabb`].
+    /// Returns the minimum distance squared to the [`Aabb`].
     /// The minimum distance is the distance to the closest point on the box,
-    /// 0 if the point is inside the box.
-    /// The maximum distance is the distance to the furthest point in the box.
+    /// or 0 if the point is inside the box.
     ///
     /// # Examples
     /// ```
@@ -610,15 +609,14 @@ impl<T: BHValue, const D: usize> Aabb<T, D> {
     ///
     /// let aabb = Aabb::with_bounds(min, max);
     /// let query = Point3::new(20.0, 0.0, 0.0);
-    /// let (min_dist, max_dist) = aabb.get_min_max_distances(&query);
-    /// assert!(min_dist == 10.0);
-    /// assert!(max_dist == (10.0_f32*10.0 + 10.0*10.0 + 20.0*20.0).sqrt());
+    /// let min_dist = aabb.get_min_distance_2(query);
+    /// assert_eq!((min_dist as f32).sqrt(), 10.0);
     /// ```
     ///
     /// [`Aabb`]: struct.Aabb.html
     ///
-    pub fn get_min_max_distances(&self, point: &Point<T, D>) -> (T, T) {
-        let half_size = self.size() * T::from_f32(0.5).unwrap();
+    pub fn get_min_distance_2(&self, point: Point<T, D>) -> T {
+        let half_size = self.half_size();
         let center = self.min + half_size;
 
         let delta = point - center;
@@ -626,19 +624,8 @@ impl<T: BHValue, const D: usize> Aabb<T, D> {
         // See <https://iquilezles.org/articles/distfunctions/>
         let q = delta.abs() - half_size;
         let outside_vec = q.map(|x| x.max(T::zero()));
-        let min_dist = outside_vec.dot(&outside_vec).sqrt(); // norm without requiring ComplexField.
 
-        // The signum helps to determine the furthest point
-        let signum = delta
-            // Invert the signum to get the furthest vertex
-            // signum never returns 0.0 so we're sure to be on a vertex.
-            .map(|x| -x.signum());
-
-        let furthest = center + signum.component_mul(&half_size);
-        let furthest_delta = point - furthest;
-        let max_dist = furthest_delta.dot(&furthest_delta).sqrt();
-
-        (min_dist, max_dist)
+        outside_vec.dot(&outside_vec)
     }
 }
 
--- a/src/aabb/aabb_impl.rs
+++ b/src/aabb/aabb_impl.rs
@@ -609,13 +609,13 @@ impl<T: BHValue, const D: usize> Aabb<T, D> {
     ///
     /// let aabb = Aabb::with_bounds(min, max);
     /// let query = Point3::new(20.0, 0.0, 0.0);
-    /// let min_dist = aabb.get_min_distance_2(query);
+    /// let min_dist = aabb.min_distance_squared(query);
     /// assert_eq!((min_dist as f32).sqrt(), 10.0);
     /// ```
     ///
     /// [`Aabb`]: struct.Aabb.html
     ///
-    pub fn get_min_distance_2(&self, point: Point<T, D>) -> T {
+    pub fn min_distance_squared(&self, point: Point<T, D>) -> T {
         let half_size = self.half_size();
         let center = self.min + half_size;
 
--- a/src/bounding_hierarchy.rs
+++ b/src/bounding_hierarchy.rs
@@ -296,7 +296,7 @@ pub trait BoundingHierarchy<T: BHValue, const D: usize> {
     /// #
     /// # impl PointDistance<f32,3> for UnitBox {
     /// #     fn distance_squared(&self, query_point: Point3<f32>) -> f32 {
-    /// #         self.aabb().get_min_distance_2(query_point)
+    /// #         self.aabb().min_distance_squared(query_point)
     /// #    }
     /// }
     /// # fn create_bvh() -> (Bvh<f32,3>, Vec<UnitBox>) {
@@ -311,14 +311,16 @@ pub trait BoundingHierarchy<T: BHValue, const D: usize> {
     ///
     /// let (bvh, shapes) = create_bvh();
     ///
-    /// let query = Point3::new(5.0, 0.0, 2.0);
-    /// let nearest_shape = bvh.nearest_from(query, &shapes);
+    /// let query = Point3::new(5.0, 5.7, 5.3);
+    /// let nearest_shape = bvh.nearest_to(query, &shapes);
+    ///
+    /// # assert_eq!(nearest_shape.unwrap().0.id, 5);
     /// ```
     ///
     /// [`BoundingHierarchy`]: trait.BoundingHierarchy.html
     /// [`Aabb`]: ../aabb/struct.Aabb.html
     ///
-    fn nearest_from<'a, Shape: BHShape<T, D> + PointDistance<T, D>>(
+    fn nearest_to<'a, Shape: BHShape<T, D> + PointDistance<T, D>>(
         &'a self,
         query: Point<T, D>,
         shapes: &'a [Shape],
@@ -344,12 +346,12 @@ impl<T: BHValue, const D: usize, H: BoundingHierarchy<T, D>> BoundingHierarchy<T
         H::traverse(self, query, shapes)
     }
 
-    fn nearest_from<'a, Shape: BHShape<T, D> + PointDistance<T, D>>(
+    fn nearest_to<'a, Shape: BHShape<T, D> + PointDistance<T, D>>(
         &'a self,
         query: Point<T, D>,
         shapes: &'a [Shape],
     ) -> Option<(&'a Shape, T)> {
-        H::nearest_from(self, query, shapes)
+        H::nearest_to(self, query, shapes)
     }
 
     fn build_with_executor<
--- a/src/bounding_hierarchy.rs
+++ b/src/bounding_hierarchy.rs
@@ -1,7 +1,8 @@
 //! This module defines the [`BoundingHierarchy`] trait.
 
 use nalgebra::{
-    ClosedAddAssign, ClosedDivAssign, ClosedMulAssign, ClosedSubAssign, Scalar, SimdPartialOrd,
+    ClosedAddAssign, ClosedDivAssign, ClosedMulAssign, ClosedSubAssign, Point, Scalar,
+    SimdPartialOrd,
 };
 use num::{Float, FromPrimitive, Signed};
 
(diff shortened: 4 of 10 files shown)


svenstaro:
Thanks a lot! I will take an in-depth look soon. How do you think this interacts with #98? I think I would like to merge #98 first and I'm almost done with it but I don't want to negate your work. Could you take a look at let me know what you think?

Azkellas:
I think the two PR are mostly orthogonal and merging #98 shouldn't affect this PR much.
Interestingly, they share some similarities in the tree traversal, but I think it's better to keep them separate as the `nearest_candidates` does not need ordering and thus gives more room for optimization. My point-bvh distance query also suffers from the fact that it needs to visit aabbs that don't contain the query point. 
This PR also doesn't provide an iterator since it uses the whole result to prune the vector (often removing ~75% of the candidates).

svenstaro:
In a related note: Would you like to help me maintain this library? It's not getting enough love and could do with some better maintenance.

svenstaro:
I merged and released #98 so this one now needs a rebase. :)

Azkellas:
I rebased without issue.
This is what the benches look like at the moment.

```
test bvh::bvh_impl::bench::bench_nearest_candidates_1200_triangles_bvh        ... bench:       1,674.28 ns/iter (+/- 82.26)
test bvh::bvh_impl::bench::bench_nearest_candidates_120k_triangles_bvh        ... bench:      15,676.73 ns/iter (+/- 838.18)
test bvh::bvh_impl::bench::bench_nearest_candidates_12k_triangles_bvh         ... bench:       4,685.55 ns/iter (+/- 165.48)
test bvh::bvh_impl::bench::bench_nearest_candidates_sponza_bvh                ... bench:      47,883.61 ns/iter (+/- 4,195.02)
test testbase::bench_nearest_candidates_120k_triangles_list                   ... bench:     664,322.50 ns/iter (+/- 39,013.88)
test testbase::bench_nearest_candidates_120k_triangles_list_aabb              ... bench:     519,004.38 ns/iter (+/- 63,937.19)
test testbase::bench_nearest_candidates_sponza_list                           ... bench:     611,325.62 ns/iter (+/- 33,850.31)
test testbase::bench_nearest_candidates_sponza_list_aabb                      ... bench:     282,300.00 ns/iter (+/- 41,833.67)
```

I'm not against becoming a co-maintainer, we can talk about it to see what needs to be done.

Azkellas:
I implemented the function for flat bvh as well. The performances are equivalent. I think the PR is ready for review now. Not perfect but functional.

> [ ] Upstream tests and benches
> [x] Implement nearest_candidate for FlatBvh
> [ ] Implement an approx version? At least test the performance gain.

Regarding the checklist in the first post, benches are here, tests are still a bit lackluster.
The approx version actually doesn't make sense since for a query point outside of the bvh bounding box it would return an empty list.

Azkellas:
Have you had a chance to take a look?

svenstaro:
Sorry, I didn't get to it yet. It's on the top of my list!

svenstaro:
@finnbear do you wanna give this a look and perhaps resolve the conflicts?

Azkellas:
Nice to see some activity on the project! I have now a faster implementation on my crate (though not public yet) that simplifies the algorithm by asking the user for more information.

This PR finds nearest candidates by only using aabbs (requiring no more information from the user), while my latest implementation add a trait to query the exact distance on relevant leaves. This allows for a way better pruning and, IIRC, makes the whole query up to 3x faster for triangles. 

I'll try to make/update a PR this week.

finnbear:
Hi :)

> I'll try to make/update a PR this week.

Does it still make sense to review & merge this PR as-is?

Azkellas:
The two versions could cohabit (nearest candidates could be useful is the cost of computing the exact distance is really high), otherwise the new version should substitute this one.
Though if the cost is high it probably means the new version would prune better too.

I think you can can wait for my update before you try to rebase this version. :)

svenstaro:
Oh, thanks for the update @Azkellas! We'll wait then. Sorry again for the long wait, life got busy. :(

Azkellas:
> Sorry again for the long wait, life got busy. :( 

No problem, I turned on my computed for the first time in three months last Wednesday, so we're quite synchronized at least. :) 

> So as I understand it, you will create a new PR? Shall we close this one then? 

I'll see if it's easier to rebase this one or upstream from scratch.

Azkellas:
I updated the code with the latest implementation.

Now instead of keeping a limit (min, max) on the distance to the best shape, we keep track of the actual distance to the current best shape (taking the first shape arbitrarily  if the bvh is not empty).

The two good news are:
- The implementation is much easier to read and maintain than before
- Performance is much much better, from 6x faster for the 1200 triangles benchmark to 50x faster for sponza (the more triangles, the faster), and 1000x faster than the most naive implementation.

The two bad news are:
- This adds the `PointDistance` trait requirement for Shape, needed to query the exact distance (squared) from the shape to the query point on candidate leaf nodes.
- The flat bvh is not optimized for distance queries as the function cannot choose which child to check first. The flat bvh is ~5-10x slower depending on the size of the bvh (still 100x faster than a naive implementation).

I think the branch is ready for review.

<details>
<summary>Incremental benchmarks</summary>

```
// naive implementation: go through each shape one by one with no BVH
test testbase::bench_nearest_candidates_120k_triangles_list                   ... bench:     681,288.75 ns/iter (+/- 55,478.50)
test testbase::bench_nearest_candidates_120k_triangles_list_aabb              ... bench:     534,145.00 ns/iter (+/- 98,617.50)
test testbase::bench_nearest_candidates_sponza_list                           ... bench:     619,938.12 ns/iter (+/- 39,642.44)
test testbase::bench_nearest_candidates_sponza_list_aabb                      ... bench:     295,907.50 ns/iter (+/- 39,118.38)
```
```
// old version with aabb min/max distances only
test bvh::bvh_impl::bench::bench_nearest_candidates_1200_triangles_bvh        ... bench:       1,744.73 ns/iter (+/- 80.47)
test bvh::bvh_impl::bench::bench_nearest_candidates_120k_triangles_bvh        ... bench:      15,921.70 ns/iter (+/- 1,739.16)
test bvh::bvh_impl::bench::bench_nearest_candidates_12k_triangles_bvh         ... bench:       4,841.70 ns/iter (+/- 399.85)
test bvh::bvh_impl::bench::bench_nearest_candidates_sponza_bvh                ... bench:      50,180.16 ns/iter (+/- 6,593.96)
```
```
// first version with exact distance query on candidate leaf nodes
test bvh::bvh_impl::bench::bench_nearest_from_1200_triangles_bvh              ... bench:       1,539.03 ns/iter (+/- 116.92)
test bvh::bvh_impl::bench::bench_nearest_from_120k_triangles_bvh              ... bench:      15,680.75 ns/iter (+/- 1,989.23)
test bvh::bvh_impl::bench::bench_nearest_from_12k_triangles_bvh               ... bench:       4,207.26 ns/iter (+/- 454.26)
test bvh::bvh_impl::bench::bench_nearest_from_sponza_bvh                      ... bench:       8,367.80 ns/iter (+/- 640.04)
```
```
// take first shape as first guess instead of T::max
test bvh::bvh_impl::bench::bench_nearest_from_1200_triangles_bvh              ... bench:       1,101.38 ns/iter (+/- 89.26)
test bvh::bvh_impl::bench::bench_nearest_from_120k_triangles_bvh              ... bench:      11,637.48 ns/iter (+/- 1,637.42)
test bvh::bvh_impl::bench::bench_nearest_from_12k_triangles_bvh               ... bench:       3,053.51 ns/iter (+/- 407.65)
test bvh::bvh_impl::bench::bench_nearest_from_sponza_bvh                      ... bench:       5,031.79 ns/iter (+/- 456.78)
```
```
// go to best min child, useful to get a better first estimation of the best child.
test bvh::bvh_impl::bench::bench_nearest_from_1200_triangles_bvh              ... bench:         276.45 ns/iter (+/- 17.75)
test bvh::bvh_impl::bench::bench_nearest_from_120k_triangles_bvh              ... bench:         776.33 ns/iter (+/- 112.84)
test bvh::bvh_impl::bench::bench_nearest_from_12k_triangles_bvh               ... bench:         426.19 ns/iter (+/- 20.88)
test bvh::bvh_impl::bench::bench_nearest_from_sponza_bvh                      ... bench:       1,211.11 ns/iter (+/- 64.11)
```

```
// comparison with flat bvh
test bvh::bvh_impl::bench::bench_nearest_from_1200_triangles_bvh              ... bench:         284.03 ns/iter (+/- 21.04)
test bvh::bvh_impl::bench::bench_nearest_from_120k_triangles_bvh              ... bench:         837.90 ns/iter (+/- 264.71)
test bvh::bvh_impl::bench::bench_nearest_from_12k_triangles_bvh               ... bench:         427.79 ns/iter (+/- 23.38)
test flat_bvh::bench::bench_nearest_from_1200_triangles_bvh                   ... bench:         700.26 ns/iter (+/- 34.75)
test flat_bvh::bench::bench_nearest_from_120k_triangles_bvh                   ... bench:       8,509.34 ns/iter (+/- 1,119.36)
test flat_bvh::bench::bench_nearest_from_12k_triangles_bvh                    ... bench:       2,148.05 ns/iter (+/- 179.71)
```
</details>

src/aabb/aabb_impl.rs
```diff
@@ -594,6 +594,39 @@ impl<T: BHValue, const D: usize> Aabb<T, D> {
...
+    ///
+    /// let aabb = Aabb::with_bounds(min, max);
+    /// let query = Point3::new(20.0, 0.0, 0.0);
+    /// let min_dist = aabb.get_min_distance_2(query);
+    /// assert_eq!((min_dist as f32).sqrt(), 10.0);
+    /// ```
+    ///
+    /// [`Aabb`]: struct.Aabb.html
+    ///
+    pub fn get_min_distance_2(&self, point: Point<T, D>) -> T {
```
finnbear, line 618: Is there a reason why we use `_2` here and `_squared` elsewhere? Perhaps they should be consistent. I personally prefer `_squared`.

src/bounding_hierarchy.rs
```diff
@@ -245,6 +247,83 @@ pub trait BoundingHierarchy<T: BHValue, const D: usize> {
...
+    /// ```
+    ///
+    /// [`BoundingHierarchy`]: trait.BoundingHierarchy.html
+    /// [`Aabb`]: ../aabb/struct.Aabb.html
+    ///
+    fn nearest_from<'a, Shape: BHShape<T, D> + PointDistance<T, D>>(
+        &'a self,
+        query: Point<T, D>,
+        shapes: &'a [Shape],
+    ) -> Option<(&'a Shape, T)>;
```
finnbear, line 325: IMO it's worth calling this in the fuzzer (`fuzz/`) for a normal BVH and a flat BVH, even if the output is disregarded, to make sure it doesn't panic. A simple and easy check would be to make sure it returns `None` if and only if the BVH is empty at that moment.

If you are unsure about this, I'm happy to implement it myself in a followup PR.

src/aabb/aabb_impl.rs
svenstaro: I agree with this. Let's keep it consistent. :)

Azkellas: Indeed, refactorization miss here

fuzz/fuzz_targets/fuzz.rs
```diff
@@ -144,6 +145,15 @@ impl<const D: usize> ArbitraryShape<D> {
     }
 }
 
+impl<const D: usize> PointDistance<Float, D> for ArbitraryShape<D> {
+    // Assume the shape is the point a.
```
svenstaro, line 149: What's point a?

svenstaro (commented):
This is very cool! @finnbear do you have time to give this a proper review? Maybe even @dashedman if you feel like it. :) I probably won't have time.

fuzz/fuzz_targets/fuzz.rs
Azkellas: It's not the case anymore, my bad, it's the center of the aabb now. Point a is one of the point that make the aabb, but before sanitization so I couldn't use it.

finnbear:
>  @finnbear do you have time to give this a proper review?

Yes! I mean, I can't start for ~3 hours, but definitely within the next couple days :)

src/bvh/bvh_impl.rs
```diff
@@ -165,6 +166,32 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
         DistanceTraverseIterator::new(self, ray, shapes)
     }
 
+    /// Traverses the [`Bvh`].
+    /// Returns the shape the closest to the query point and the distance to it.
+    ///
```
finnbear, line 171: Wording feels slightly awkward, recommend: "Returns the closest shape to..."

src/bvh/bvh_impl.rs
```diff
@@ -165,6 +166,32 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
...
+    pub fn nearest_from<'a, Shape: Bounded<T, D> + PointDistance<T, D>>(
+        &self,
+        origin: nalgebra::Point<T, D>,
+        shapes: &'a [Shape],
+    ) -> Option<(&'a Shape, T)>
+    where
+        Self: std::marker::Sized,
+    {
+        if shapes.is_empty() {
```
finnbear, line 184: To be safe, considering the caller can manipulate `shapes`, let's use `self.nodes.is_empty()` instead.

src/bvh/bvh_impl.rs
```diff
@@ -165,6 +166,32 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
...
+            return None;
+        }
+
+        let mut best_candidate = (&shapes[0], shapes[0].distance_squared(origin));
```
finnbear, line 188: Furthermore, `shapes[0]` is not *technically* guaranteed to be in the BVH. The user may have removed it without actually swapping/clearing it from the slice. I recommend passing down an `&mut Option<&Shape, T>` or something similar which can later be `unwrap`'d.

src/bvh/bvh_node.rs
```diff
@@ -323,6 +324,57 @@ impl<T: BHValue, const D: usize> BvhNode<T, D> {
...
         }
     }
+
+    /// Traverses the [`Bvh`] recursively and updates the given `best_candidate` with
+    /// the nearest shape found so far.
+    ///
+    /// [`Aabb`]: ../aabb/struct.Aabb.html
+    /// [`Bvh`]: struct.Bvh.html
+    ///
+    pub fn nearest_from_recursive<'a, Shape: Bounded<T, D> + PointDistance<T, D>>(
```
finnbear, line 334: I realize that other similar functions may still be public, but AFAIK we're moving away from doing that. Let's make this new one private or `pub(crate)`, as needed.

src/bvh/bvh_node.rs
```diff
@@ -323,6 +324,57 @@ impl<T: BHValue, const D: usize> BvhNode<T, D> {
...
+        nodes: &[BvhNode<T, D>],
+        node_index: usize,
+        origin: nalgebra::Point<T, D>,
```
finnbear, line 337: IMO `query` might be clearer than `origin`

src/flat_bvh.rs
```diff
@@ -419,6 +421,131 @@ impl<T: BHValue + std::fmt::Display, const D: usize> BoundingHierarchy<T, D> for
...
+    fn nearest_from<'a, Shape: BHShape<T, D> + PointDistance<T, D>>(
+        &'a self,
+        query: Point<T, D>,
+        shapes: &'a [Shape],
+    ) -> Option<(&'a Shape, T)> {
+        if self.is_empty() {
+            return None;
+        }
+
+        let mut best_element = (&shapes[0], shapes[0].distance_squared(query));
```
finnbear, line 510: `shapes[0]` may suffer from the same issue as in the normal BVH (see previous comment)

src/testbase.rs
```diff
@@ -718,3 +864,122 @@ pub fn intersect_12k_triangles_bh<T: BoundingHierarchy<f32, 3>>(b: &mut ::test::
...
+    triangles: &[Triangle],
+    bounds: &TAabb3,
+    b: &mut ::test::Bencher,
+) {
+    let mut seed = 0;
+    b.iter(|| {
+        let point = next_point3(&mut seed, bounds);
+
+        // Traverse the [`BoundingHierarchy`] recursively.
+        let _best_result = bh.nearest_from(point, triangles);
```
finnbear, line 956: This benchmark could be optimized away unless you pass the output to `std::hint::black_box`.

Also applies to:
- `nearest_from_list`
- `nearest_from_list_aabb`

finnbear (changes requested):
Thanks for these changes! :rocket:

I've left some feedback that I hope you can address (if not, that's fine; I will personally drive the PR to completion rather than letting it become too stale).

As a final note, to the extent you feel comfortable doing so, please `git rebase` your revisions into meaningful commits. You can err on the side of squashing commits, especially those that "fix" prior commits. You may find the `git absorb` tool useful.

fuzz/fuzz_targets/fuzz.rs
```diff
@@ -144,6 +145,14 @@ impl<const D: usize> ArbitraryShape<D> {
     }
 }
 
+impl<const D: usize> PointDistance<Float, D> for ArbitraryShape<D> {
+    fn distance_squared(&self, query_point: Point<Float, D>) -> Float {
+        let center = self.aabb().center();
+        let offset = query_point - center;
+        offset.dot(&offset)
```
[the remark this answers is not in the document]
Azkellas: It's two different things.
The `Aabb::min_distance_squared` is used to prune for cheap in the tree traversal with a low estimation of the actual distance, while `PointDistance` is the (potentially expensive) actual distance on leaf nodes.

Here I decided in the fuzzer the objects were points, in the center of the cell. I could have decided the objects were the aabb themselves, but I prefered having two different objects to consider. Ideally the object should be a random point/shape in the aabb.

fuzz/fuzz_targets/fuzz.rs
```diff
@@ -90,7 +91,7 @@ struct ArbitraryShape<const D: usize> {
 
 impl<const D: usize> Debug for ArbitraryShape<D> {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
-        Debug::fmt(&self.aabb(), f)
+        Debug::fmt(&(&self.aabb(), &self.a), f)
```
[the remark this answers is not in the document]
Azkellas: There was when the actual object was `a` and not the center of the aabb, but that didn't work as `a` was often outside `self.aabb()` due to how it's generated.

fuzz/fuzz_targets/fuzz.rs
```diff
@@ -317,6 +326,70 @@ impl<const D: usize> Workload<D> {
...
+    ///
+    /// # Panics
+    /// If `assert_agreement` is true, panics if the results differ in an
+    /// unexpected way.
+    fn fuzz_nearest_from<'a>(
+        &'a self,
+        bvh: &'a Bvh<Float, D>,
+        flat_bvh: &'a FlatBvh<Float, D>,
+        query_point: Point<Float, D>,
+        assert_agreement: bool,
```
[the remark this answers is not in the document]
Azkellas: Yeah tbh I used the assert agreement to match the traversal signature without looking at how it was called. Indeed it doesn't really makes sense in that case I think.

I think we should remove the boolean but keep all current asserts, even if they're not perfect (e.g. `float_cmp::approx_eq`) they are still a safeguard if things go really wrong.

src/bvh/bvh_impl.rs
Azkellas: would it makes sense to use `self.nodes[0]` instead of `shapes[0]` or even that is not guaranteed?

finnbear: Sure, you would then have to handle both cases (`self.nodes[0]` is an internal node **or** a leaf node)

src/testbase.rs
Azkellas: Not sure where you want to put them for the list/aabb versions, unless you want to wrap the call to the functions themselves.

Azkellas:
I adressed the review and rebased to latest main commit. A couple points are still open.
Regarding the main comment and squashing, I tend to squash as I go in PRs, so I didn't change the number/name of the commit during this review, tell me if you want less commits.

src/testbase.rs
finnbear: You could pass `min_dist` to `black_box` at the end :)

svenstaro:
I prefer not squashing so long as the commits in a PR tell a story and are properly rebased. [git-absorb](https://github.com/tummychow/git-absorb) is quite helpful here.

src/flat_bvh.rs
```diff
@@ -419,6 +421,134 @@ impl<T: BHValue + std::fmt::Display, const D: usize> BoundingHierarchy<T, D> for
...
+        // The traversal loop should terminate when `max_length` is set as the next node index.
+        let max_length = self.len();
+
+        // Iterate while the node index is valid.
+        while index < max_length {
+            let node = &self[index];
+
+            if node.entry_index == u32::MAX {
+                // If the entry_index is MAX_UINT32, then it's a leaf node.
+                let shape = &shapes[node.shape_index as usize];
```
dashedman, line 526: Maybe we should move this rule to function `fn is_leaf(&self) -> bool` ?

finnbear: Good idea!

dashedman:
So, what are you doing here?
 New feature just returns nearest shape to the any passed point?

src/bvh/bvh_impl.rs
```diff
@@ -472,6 +498,14 @@ impl<T: BHValue + std::fmt::Display, const D: usize> BoundingHierarchy<T, D> for
         self.traverse(query, shapes)
     }
 
+    fn nearest_from<'a, Shape: BHShape<T, D> + PointDistance<T, D>>(
+        &'a self,
+        query: nalgebra::Point<T, D>,
+        shapes: &'a [Shape],
+    ) -> Option<(&'a Shape, T)> {
+        self.nearest_from(query, shapes)
```
dashedman, line 506: I don't understand. It's just an infinite recursive or what?

src/bvh/bvh_node.rs
```diff
@@ -323,6 +324,57 @@ impl<T: BHValue, const D: usize> BvhNode<T, D> {
...
+    ///
+    /// [`Aabb`]: ../aabb/struct.Aabb.html
+    /// [`Bvh`]: struct.Bvh.html
+    ///
+    pub(crate) fn nearest_from_recursive<'a, Shape: Bounded<T, D> + PointDistance<T, D>>(
+        nodes: &[BvhNode<T, D>],
+        node_index: usize,
+        query: nalgebra::Point<T, D>,
+        shapes: &'a [Shape],
+        best_candidate: &mut Option<(&'a Shape, T)>,
```
dashedman, line 339: It's just my tastes, but I'm prefer to return new value from func instead of using in-out args

src/bvh/bvh_impl.rs
finnbear: This is within
```
impl<T: BHValue + std::fmt::Display, const D: usize> BoundingHierarchy<T, D> for Bvh<T, D> {
```

A trait function is being implemented by calling an inherent function of the same name.

dashedman: Why are we not write implementation just in trait imply for this structure?

finnbear: I would assume:
1.  consistency. prior code uses both inherent and trait functions
2. usability. using a trait function requires importing the trait; using an inherent function doesn't

dashedman: Ok

src/bvh/bvh_node.rs
svenstaro: I agree with this. The rest of the library also uses return values.

svenstaro (changes requested):

src/bvh/bvh_impl.rs
Azkellas: It was done for consistency with the traverse methods.

src/bvh/bvh_node.rs
Azkellas: While it's possible to use return values to update best_candidate, I'm not sure it makes the code much cleaner. Worse, the performances drop.
I checked both versions traverse the tree the same way to make sure it's not a logic issue, so I guess it's a compiler optimization. I haven't investigated more.

with `&mut`
```
test bvh::bvh_impl::bench::bench_nearest_from_1200_triangles_bvh              ... bench:         284.00 ns/iter (+/- 20.23)
test bvh::bvh_impl::bench::bench_nearest_from_120k_triangles_bvh              ... bench:         812.96 ns/iter (+/- 169.83)
test bvh::bvh_impl::bench::bench_nearest_from_12k_triangles_bvh               ... bench:         427.06 ns/iter (+/- 23.12)
test bvh::bvh_impl::bench::bench_nearest_from_sponza_bvh                      ... bench:       1,216.93 ns/iter (+/- 92.07)
```

when returning best candidate

```
test bvh::bvh_impl::bench::bench_nearest_from_1200_triangles_bvh              ... bench:         396.98 ns/iter (+/- 16.86)
test bvh::bvh_impl::bench::bench_nearest_from_120k_triangles_bvh              ... bench:       1,033.95 ns/iter (+/- 402.01)
test bvh::bvh_impl::bench::bench_nearest_from_12k_triangles_bvh               ... bench:         577.19 ns/iter (+/- 44.53)
test bvh::bvh_impl::bench::bench_nearest_from_sponza_bvh                      ... bench:       1,474.12 ns/iter (+/- 170.15)
```

<details>
<summary>Code to test with return value</summary>

```rust

    pub(crate) fn nearest_from_recursive<'a, Shape: Bounded<T, D> + PointDistance<T, D>>(
        nodes: &[BvhNode<T, D>],
        node_index: usize,
        query: nalgebra::Point<T, D>,
        shapes: &'a [Shape],
        mut best_candidate: Option<(&'a Shape, T)>,
    ) -> Option<(&'a Shape, T)> {
        match nodes[node_index] {
            BvhNode::Node {
                ref child_l_aabb,
                child_l_index,
                ref child_r_aabb,
                child_r_index,
                ..
            } => {
                // Compute the min dist for both children
                let mut children = [
                    (child_l_index, child_l_aabb.min_distance_squared(query)),
                    (child_r_index, child_r_aabb.min_distance_squared(query)),
                ];

                // Sort children to go to the best candidate first and have a better chance of pruning
                if children[0].1 > children[1].1 {
                    children.swap(0, 1);
                }

                // Traverse children
                for (index, child_dist) in children {
                    // Node might contain a better shape: check it.
                    if best_candidate.is_none_or(|(_, best_dist)| child_dist <= best_dist) {
                        best_candidate = Self::nearest_from_recursive(
                            nodes,
                            index,
                            query,
                            shapes,
                            best_candidate,
                        );
                    }
                }
            }
            BvhNode::Leaf { shape_index, .. } => {
                // This leaf might contain a better shape: check it directly with its exact distance (squared).
                let dist = shapes[shape_index].distance_squared(query);

                if best_candidate.is_none_or(|(_, best_dist)| dist < best_dist) {
                    best_candidate = Some((&shapes[shape_index], dist));
                }
            }
        }

        best_candidate
    }

````
</details>

Comment (svenstaro) [reply src/bvh/bvh_node.rs:339]: Alright, then let's keep your original version.

Comment (dashedman) [reply src/bvh/bvh_node.rs:339]: Try please this version for me 

```rust
pub(crate) fn nearest_from_recursive<'a, Shape: Bounded<T, D> + PointDistance<T, D>>(
        nodes: &[BvhNode<T, D>],
        node_index: usize,
        query: nalgebra::Point<T, D>,
        shapes: &'a [Shape],
        best_candidate: Option<(&'a Shape, T)>,
    ) -> Option<(&'a Shape, T)> {
        match nodes[node_index] {
            BvhNode::Node {
                ref child_l_aabb,
                child_l_index,
                ref child_r_aabb,
                child_r_index,
                ..
            } => {
                // Compute the min dist for both children
                let mut children = [
                    (child_l_index, child_l_aabb.min_distance_squared(query)),
                    (child_r_index, child_r_aabb.min_distance_squared(query)),
                ];

                // Sort children to go to the best candidate first and have a better chance of pruning
                if children[0].1 > children[1].1 {
                    children.swap(0, 1);
                }

                // Traverse children
                for (index, child_dist) in children {
                    // Node might contain a better shape: check it.
                    if best_candidate.is_none_or(|(_, best_dist)| child_dist <= best_dist) {
                        return Self::nearest_from_recursive(
                            nodes,
                            index,
                            query,
                            shapes,
                            best_candidate,
                        )
                    }
                };
                best_candidate
            }
            BvhNode::Leaf { shape_index, .. } => {
                // This leaf might contain a better shape: check it directly with its exact distance (squared).
                let dist = shapes[shape_index].distance_squared(query);

                if best_candidate.is_none_or(|(_, best_dist)| dist < best_dist) {
                    Some((&shapes[shape_index], dist))
                }
            }
        }
    }
```

Azkellas:
Cargo fuzz does not build with latest nightly but that's not on me. Otherwise I think I dealt with all the reviews, I let you double check and close them to be sure I didn't miss any.

finnbear:
> Cargo fuzz does not build with latest nightly but that's not on me.

Sorry about that! I just merged #143 which fixes the CI to a known-good version of nightly. If you rebase on to `master`, you should get the benefit of that.

fuzz/fuzz_targets/fuzz.rs
```diff
@@ -317,6 +326,56 @@ impl<const D: usize> Workload<D> {
         traverse
     }
 
+    /// Compares normal and flat-BVH nearest_from query.
+    ///
+    /// # Panics
+    /// If `assert_agreement` is true, panics if the results differ in an
+    /// unexpected way.
```
finnbear, line 333: This comment should no longer reference `assert_agreement`

fuzz/fuzz_targets/fuzz.rs
```diff
@@ -319,6 +328,56 @@ impl<const D: usize> Workload<D> {
...
+        bvh: &'a Bvh<Float, D>,
+        flat_bvh: &'a FlatBvh<Float, D>,
+        query_point: Point<Float, D>,
+    ) {
+        let best = bvh
+            .nearest_from(query_point, &self.shapes)
+            .map(|(s, d)| (ByPtr(s), d));
+        let flat_best = flat_bvh
+            .nearest_from(query_point, &self.shapes)
+            .map(|(s, d)| (ByPtr(s), d));
```
finnbear, line 347: I think the `ByPtr` wrapper isn't necessary here (or later in this function) because, as you say, you don't want to compare the shapes.

fuzz/Cargo.toml
```diff
@@ -10,6 +10,7 @@ cargo-fuzz = true
 
 [dependencies]
 arbitrary = { version = "1.4.1", features = ["derive"] }
+float-cmp = "0.10.0"
```
finnbear, line 13: `bvh` itself uses https://github.com/svenstaro/bvh/blob/ca417859cd3f33b129cd82079d1a90d7da543e0a/Cargo.toml#L18

```
abs_diff_eq!(1.0, 1.0);
abs_diff_eq!(1.0, 1.0, epsilon = f64::EPSILON);
```

Maybe the fuzzer should too, to avoid dependency proliferation.

src/flat_bvh.rs
```diff
@@ -41,6 +43,14 @@ pub struct FlatNode<T: BHValue, const D: usize> {
     pub shape_index: u32,
 }
 
+impl<T: BHValue, const D: usize> FlatNode<T, D> {
+    /// Returns whether the node is a leaf node.
+    #[must_use]
+    fn is_leaf(&self) -> bool {
+        self.entry_index == u32::MAX
+    }
```
finnbear, line 51: I'd prefer `pub fn`, as long as the rest of the `FlatNode` internals are public.

finnbear (changes requested):
Looking good to me :+1:

I had a few more comments (see above).

Also, a slightly more subjective concern: would the name `nearest_to` be better than `nearest_from`? Even your documentation uses "to" in "Returns the shape closest **to** the query point".

src/bounding_hierarchy.rs
```diff
@@ -245,6 +247,85 @@ pub trait BoundingHierarchy<T: BHValue, const D: usize> {
         shapes: &'a [Shape],
     ) -> Vec<&'a Shape>;
 
+    /// Traverses the [`BoundingHierarchy`].
+    /// Returns the shape closest to the query point.
```
finnbear, line 251: A subjective concern: would the name `nearest_to` be better than `nearest_from`? Your documentation uses "to" here ;)

Azkellas:
Updated.

It has to be noted that the CI still fails because I used [Option::is_none_or](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_none_or) that was introduced in 1.82.0. I don't know if bvh has a MSRV policy nor how it should evolved with the new cargo resolver. Let me know if that's an issue.

finnbear:
Thanks! I'm not aware of an MSRV policy but, out of self interest (I use a nightly compiler from almost a year ago, and upgrade it annually, based on how long it takes to find a specific nightly that compiles all my projects), it would be nice for the crate to work on Rust versions older than ~3 months.

Removing `is_none_or` may, however, require silencing a `clippy` lint, which is also undesirable.

I did a quick survey of MSRV's of other crates:
- `tokio`: currently >1 year old, but only guaranteed to be >=6 months old
- `regex`: currently 2-3 years old
- `glam`: currently >1 year old
- `nalgebra`: not explicitly listed

1y seems like a common denominator to aim for.

I suggest removing it, silencing the `clippy` lint that probably results from that, but adding a clear `TODO` comment to switch to `is_none_or` after Oct 17, 2025 (when `is_none_or` will have been stable for a year).

Azkellas:
Funnily enough the clippy lint doesn't work when dereferencing a reference to an option, but I had to use `#[allow(...)]` instead of `#[expect(...)]` since `expect` is too recent as well. The CI doesn't complain anymore.

finnbear (approved):
Nice work!

PR #114, opened 2024-12-29
Fix all clippy warnings.

finnbear:
This PR fixes all clippy warnings, that currently affect my other PR #113 :)

### DIFF
--- a/src/bvh/iter.rs
+++ b/src/bvh/iter.rs
@@ -107,6 +107,7 @@ impl<'bvh, 'shape, T: BHValue, const D: usize, Shape: Bounded<T, D>>
     }
 }
 
+#[allow(clippy::needless_lifetimes)]
 impl<'bvh, 'shape, T: BHValue, const D: usize, Shape: Bounded<T, D>> Iterator
     for BvhTraverseIterator<'bvh, 'shape, T, D, Shape>
 {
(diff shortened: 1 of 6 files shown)


src/bvh/iter.rs
```diff
@@ -107,6 +107,7 @@ impl<'bvh, 'shape, T: BHValue, const D: usize, Shape: Bounded<T, D>>
     }
 }
 
+#[allow(clippy::needless_lifetimes)]
 impl<'bvh, 'shape, T: BHValue, const D: usize, Shape: Bounded<T, D>> Iterator
     for BvhTraverseIterator<'bvh, 'shape, T, D, Shape>
```
svenstaro, line 112: Can't we do this instead?
```suggestion
impl<'shape, T: BHValue, const D: usize, Shape: Bounded<T, D>> Iterator
    for BvhTraverseIterator<'_, 'shape, T, D, Shape>
```

svenstaro (changes requested):
Nice! Just curious whether you opted for `#[allow(clippy::needless_lifetimes)]` as opposed to going with the way clippy suggests to fix this.

src/bvh/iter.rs
finnbear: Yes, that works too! I originally chose `#[allow(clippy::needless_lifetimes)]` because I didn't know whether the needless lifetime with a special name "bvh" was unintentional or intentional (for consistent and self-documentation).

svenstaro: I _do_ like the consistency of it but I think `#[allow()]` things are really a last resort. I think this case is really 50/50 but let's go with clippy's suggestion.

svenstaro:
Good stuff.

PR #129, opened 2025-01-08
Fix #124 - fix typo in assertion.

### LINKED ISSUE #124: Removing either shape from a 2-shape BVH panics
The following code panics but shouldn't:
```
    #[test]
        let mut boxes = vec![
            UnitBox::new(0, TPoint3::new(0.0, 0.0, 0.0)),
            UnitBox::new(1, TPoint3::new(5.0, 5.0, 5.0)),
        ];
        let mut bvh = Bvh::build(&mut boxes);
        bvh.remove_shape(&mut boxes, 0, true); // removing 1 also panics
    }
```
```
thread 'bvh::distance_traverse::tests::test_traverse_subtle_bvh' panicked at src/bvh/optimization.rs:250:17:
Circular node that wasn't root parent=0 node=1
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a4cb3c831823d9baa56c3d90514b75b2660116fa/library/std/src/panicking.rs:681:5
   1: core::panicking::panic_fmt
             at /rustc/a4cb3c831823d9baa56c3d90514b75b2660116fa/library/core/src/panicking.rs:75:14
   2: bvh::bvh::optimization::<impl bvh::bvh::bvh_impl::Bvh<T,_>>::remove_shape
             at ./src/bvh/optimization.rs:250:17
   3: bvh::bvh::distance_traverse::tests::test_traverse_subtle_bvh
             at ./src/bvh/distance_traverse.rs:386:9
   4: bvh::bvh::distance_traverse::tests::test_traverse_subtle_bvh::{{closure}}
             at ./src/bvh/distance_traverse.rs:380:34
   5: core::ops::function::FnOnce::call_once
             at /home/finnb/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
   6: core::ops::function::FnOnce::call_once
             at /rustc/a4cb3c831823d9baa56c3d90514b75b2660116fa/library/core/src/ops/function.rs:250:5
```

finnbear:
I'm starting to understand the implementation a bit more. I can now see that the assertion asserted exactly the opposite of what was intended. Fixing that allows the fuzzer to remove nodes without panics or inconsistencies.

Fixes #124


src/bvh/optimization.rs
```diff
@@ -288,9 +298,7 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
             if deleted_shape_index < end_shape {
                 shapes.swap(deleted_shape_index, end_shape);
                 let node_index = shapes[deleted_shape_index].bh_node_index();
-                if let Some(index) = self.nodes[node_index].shape_index_mut() {
-                    *index = deleted_shape_index
-                }
+                *self.nodes[node_index].shape_index_mut().unwrap() = deleted_shape_index;
```
svenstaro, line 301: Is it safe to unwrap this? Are we absolutely sure this can never crash? If so, please comment why this is the case. If there's a remote possibility that it might crash, use `expect()`.

src/bvh/optimization.rs
```diff
@@ -324,12 +332,13 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
                     let l_aabb = self.nodes[child_l_index].get_node_aabb(shapes);
                     let r_aabb = self.nodes[child_r_index].get_node_aabb(shapes);
                     let mut stop = true;
-                    let epsilon = T::from_f32(0.00001).unwrap_or(T::zero());
-                    if !l_aabb.relative_eq(&child_l_aabb, epsilon) {
+                    // Avoid `relative_eq`, because rounding errors can accumulate and
+                    // eventually the BVH won't necessarily be tight.
+                    if l_aabb != child_l_aabb {
```
svenstaro, line 337: Good idea. I wonder why this wasn't done before.

src/bvh/bvh_impl.rs
```diff
@@ -377,7 +377,9 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
         } = self.nodes[node_index]
         {
             let joint_aabb = child_l_aabb.join(&child_r_aabb);
-            assert!(joint_aabb.relative_eq(outer_aabb, T::epsilon()));
+            // Avoid `relative_eq`, since there is no reason for imprecision, since
+            // Aabb's are joined by selecting the min/max of coordinates.
```
svenstaro, line 381: The comment doesn't really make sense without the context of this change (the reader won't know it was asserting with `relative_eq` before) I think. Perhaps instead explain why it's correct to do an exact comparison.

src/bvh/optimization.rs
```diff
@@ -354,19 +363,13 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
...
-                self.nodes.truncate(end);
-                return;
-            }
             let parent = self.nodes[parent_index];
+            assert!(!parent.is_leaf());
             let moved_left = parent.child_l() == end;
-            if !moved_left && parent.child_r() != end {
-                self.nodes.truncate(end);
-                return;
-            }
```
svenstaro, line 366: Why is it safe to take this out? I admit my understanding of the implementation is a bit rusty at this point.

svenstaro (changes requested):
Top notch working finding the problem! Just some small nits. :)

src/bvh/optimization.rs
finnbear: If the **parent** of a node is a **leaf** node, the BVH is inconsistent. 75%+ of the code will panic on an inconsistent BVH, as it should. It shouldn't be possible for the BVH to become inconsistent. Trying to handle cases that should be impossible in a small minority of places isn't a solution to bugs that make the BVH inconsistent. The fuzzer is supposed to catch all such bugs in advance ;)

src/bvh/optimization.rs
finnbear: `Option::unwrap` is analogous to `slice[index]`. They'll both panic if expectations are violated (the BVH is inconsistent, which should never happen). To be clear, are you suggesting to only comment the occasional `unwrap` but not the very common `[..]`?

I can certainly comment this, but it feels inconsistent.

svenstaro: That's a good point actually. Perhaps we are in fact too liberal with `[]` where we should be using `.get()` with a proper error? I don't currently have a good suggestion here since we use `[]` a whole lot. At any rate, this works for now. :)

svenstaro:
Good work, merging.

PR #128, opened 2025-01-07
Add generic traversal

finnbear:
## Changes
- [x] Make traversal (normal and iterator, but not distance-based) generic **(breaking change)**
  - [x] Normal BVH
  - [x] Flat BVH
  - [x] `BoundingHierarchy` trait
  - [x] Traverse BVH by ray (existing feature maintained)
  - [x] Traverse BVH by point
  - [x] Traverse BVH by AABB
  - [x] Traverse BVH by ball (2d = circle, 3d = sphere)
- [x] Fuzz
- [x] Test
  - [x] `Ball::contains`
  - [x] `Aabb::intersects_aabb`
- [x] Documentation

## Questions
- [x] Would `IntersectsAabb` be a better name for `AabbIntersection`?
- [x] Is `Query` a good name for the `IntersectsAabb` generic?

## Related
Fixes #42
Fixes #45
Supersedes #43 (credit to @medakk for the central idea of this PR)
Related #71

PR #130, opened 2025-01-08
Fix #120 - traverse single-node BVH

### LINKED ISSUE #120: `Bvh::traverse` produces incorrect results
The following test case fails but should pass:
```rust
    #[test]
        let mut boxes = vec![
            UnitBox::new(0, TPoint3::new(0.0, 1.0, 2.0)),
        ];
        let ray = Ray::new(
            TPoint3::new(0.0, 0.0, 0.0),
            SVector::<f32, 3>::new(1.0, 0.0, 0.0),
        );

        let bvh = Bvh::build(&mut boxes);

        let found = bvh.traverse(&ray, &boxes);
        assert_eq!(found.len(), 0);
    }
```
The ray trivially misses the unit-cube:

Screenshot from 2025-01-06 20-23-16

finnbear:
Single-node BVH's are a special case since the root node is a leaf node. All traversal types, except flat, were unconditionally returning leaf nodes that they processed because they incorrectly assumed only intersected child leaf would be processed.

Fixes #120 and asserts, via the fuzzer, that traversal (which is now correct) agrees with flat traversal (which was correct).

Superficially conflicts with #128 and that PR is bigger, so recommend merging that PR first to make it easier to fix conflicts.

### DIFF
--- a/src/bvh/bvh_impl.rs
+++ b/src/bvh/bvh_impl.rs
@@ -107,7 +107,7 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
             return Vec::new();
         }
         let mut indices = Vec::new();
-        BvhNode::traverse_recursive(&self.nodes, 0, ray, &mut indices);
+        BvhNode::traverse_recursive(&self.nodes, 0, shapes, ray, &mut indices);
         indices
             .iter()
             .map(|index| &shapes[*index])
@@ -454,7 +454,13 @@ pub fn rayon_executor<S, T: Send + BHValue, const D: usize>(
 
 #[cfg(test)]
 mod tests {
-    use crate::testbase::{build_empty_bh, build_some_bh, traverse_some_bh, TBvh3, TBvhNode3};
+    use crate::{
+        bounding_hierarchy::BoundingHierarchy,
+        testbase::{
+            build_empty_bh, build_some_bh, traverse_some_bh, TBvh3, TBvhNode3, TPoint3, TRay3,
+            TVector3, UnitBox,
+        },
+    };
 
     #[test]
     /// Tests whether the building procedure succeeds in not failing.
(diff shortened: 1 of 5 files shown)


src/bvh/iter.rs
```diff
@@ -33,7 +33,14 @@ impl<'bvh, 'shape, T: BHValue, const D: usize, Shape: Bounded<T, D>>
...
             stack_size: 0,
-            has_node: !bvh.nodes.is_empty(),
+            has_node: match bvh.nodes.first() {
+                // Only process the root leaf node if the shape's AABB is intersected.
+                Some(BvhNode::Leaf { shape_index, .. }) => {
+                    ray.intersects_aabb(&shapes[*shape_index].aabb())
+                }
+                Some(_) => true,
+                None => false,
+            },
```
svenstaro, line 43: Please pull this out of the construction statement and just place it above and store the value in a `has_node` variable. Should make it a bit more readable. I don't like it when the construction itself has too much logic in it. The previous logic here was fine because it wasn't much code.

src/bvh/iter.rs
```diff
@@ -33,7 +33,14 @@ impl<'bvh, 'shape, T: BHValue, const D: usize, Shape: Bounded<T, D>>
             stack: [0; 32],
             node_index: 0,
             stack_size: 0,
-            has_node: !bvh.nodes.is_empty(),
+            has_node: match bvh.nodes.first() {
+                // Only process the root leaf node if the shape's AABB is intersected.
```
svenstaro, line 37: Maybe provide some reasoning here _why_ this is a good idea (even if it seems obvious to you).

src/bvh/distance_traverse.rs
```diff
@@ -50,7 +50,14 @@ where
...
             stack_size: 0,
-            has_node: !bvh.nodes.is_empty(),
+            has_node: match bvh.nodes.first() {
+                // Only process the root leaf node if the shape's AABB is intersected.
+                Some(BvhNode::Leaf { shape_index, .. }) => {
+                    ray.intersects_aabb(&shapes[*shape_index].aabb())
+                }
+                Some(_) => true,
+                None => false,
+            },
```
svenstaro, line 60: See comments in `src/bvh/iter.rs`.

src/bvh/bvh_impl.rs
```diff
@@ -107,7 +107,17 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
             return Vec::new();
         }
         let mut indices = Vec::new();
-        BvhNode::traverse_recursive(&self.nodes, 0, ray, &mut indices);
+        // Manually handle the case of a single-node BVH, because `traverse_recursive`
+        // cannot access the root shape's AABB.
```
svenstaro, line 111: We _could_ change this I suppose and give `traverse_recursive()` a reference to the root _in theory_. I'm not sure it would be cleaner in the end. What do you think? We could even give it `&self` and make the `BvhNode` carry a reference to the root (though this would probably be a large change and I'm not sure it's even useful).

svenstaro (changes requested):

src/bvh/bvh_impl.rs
finnbear: I would have done that except for the fact that `traverse_recursive` is public-API. AFAIK it shouldn't be. If you're fine with privatizing it, I'd make this change.

svenstaro: That's a good point. I don't see a reason people should use `traverse_recursive` directly over `traverse`.

src/bvh/iter.rs
finnbear: Since this logic was duplicated, and I didn't want to also duplicate the comment, I abstracted this into a function.

src/bvh/distance_traverse.rs
finnbear: Made a function to handle both :)

svenstaro:
Man, you're really cleaning house. :)

PR #131, opened 2025-01-09
Bugfix #107 - intersect no depth AABB

### LINKED ISSUE #107: Null ray intersect with a flat AABB
Hello,

thank you for this useful crate :) It works great for my application, except for the following case. When intersecting a ray and a flat AABB, `false` is returned. For instance,

```rust
    let origin = Point3::<f64>::new(0.0, 0.0, 0.0);
    let direction = Vector3::<f64>::new(0.0, 0.0, 1.0);
    let ray = Ray::new(origin, direction);
    let min = Point3::<f64>::new(-1.0, -1.0, 1.0);
    let max = Point3::<f64>::new(1.0, 1.0, 1.0);
    let aabb = Aabb::with_bounds(min, max);
    assert!(ray.intersects_aabb(&aabb));
```

The last assertion fails, where I would have expected `true` (e.g. when bounding an axis-aligned triangular facet).

My apologies if this is the intended behaviour?

I am running on x86_64 Linux with version 0.9.0 (and default features, i.e. `simd` disabled).

finnbear:
Fixes #107

Not only does the fix proposed in #107 work, it is intuitively correct.


svenstaro:
Funny how these things went unfixed for so long :) Good fix!

PR #132, opened 2025-01-10
Fix #121 - add justfile.

### LINKED ISSUE #121: Add a justfile
Since #113, the number of useful commands has grown and a justfile would be useful:

- testing
- benchmarking
- fmt both `src/` and `fuzz/`
- clippy both `src/` and `fuzz/`
- fuzzing (`cargo fuzz run fuzz`)

finnbear:
Fixes #121

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,0 +1,38 @@
+# quick sanity check, useful to run before committing
+lint: fmt check clippy
+
+# format src and fuzzer
+fmt:
+    cargo fmt
+    cargo fmt --manifest-path fuzz/Cargo.toml
+
+# type-check src and the fuzzer
+#
+# Type-checking with and without features helps catch,
+# for example, an import statement that is `#[cfg(_)]`'d
+# to a different feature than the code that relies on it.
+check:
+    cargo check
+    cargo check --all-features
+    cargo check --manifest-path fuzz/Cargo.toml
+
+# run clippy on src and fuzzer
+clippy:
+    cargo clippy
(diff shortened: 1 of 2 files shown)


justfile
```diff
@@ -0,0 +1,34 @@
...
+
+# format src and fuzzer
+fmt:
+    cargo fmt
+    cargo fmt --manifest-path fuzz/Cargo.toml
+
+# type-check with and without features, and the fuzzer
+check:
+    cargo check
+    cargo check --all-features
```
svenstaro, line 12: Isn't it sufficient to just have `cargo check --all-features`?

justfile
```diff
@@ -0,0 +1,34 @@
+# useful to run when submitting a PR
+pr: fmt check clippy
```
svenstaro, line 2: I usually call this target `lint`.

svenstaro (changes requested):
You should also describe its usage in the `README` as otherwise most people won't know what to do with it.

Otherwise this looks solid to me.

justfile
finnbear: It is possible for the code to work with `--all-features` but not with default features, such as when a `use` statement required for default features was accidentally `#[cfg(...)]`'d to a non-default feature.

svenstaro: Ah, that makes sense. In that case, perhaps add a comment explaining that so it's not removed in the future.

finnbear:
Thanks for the review; done!

svenstaro:
That works. Merging as-is.

PR #133, opened 2025-01-10
Bugfix #119 - distance traversal don't return non-hit shapes

### LINKED ISSUE #119: `DistanceTraverseIterator` produces incorrect results
This test case fails but should succeed.
```
    #[test]
        let mut boxes = vec![
            // Neither AABB should be intersected.
            UnitBox::new(0, TPoint3::new(-50.0, -50.0, -25.0)),
            UnitBox::new(1, TPoint3::new(0.0, 0.0, 0.0)), // problem goes away if this one is removed.
        ];
        let ray = TRay3::new(
            TPoint3::new(-50.0, -50.0, -50.0),
            TVector3::new(1.0, 0.0, 0.0),
        );

        let bvh = TBvh3::build(&mut boxes);

        let found = bvh.nearest_traverse_iterator(&ray, &boxes).next();
        assert!(found.is_none(), "should not have found id={:?}", found.unwrap().id);
    }
```
```
should not have found id=0
```

finnbear:
The old implementation of `Ray::intersection_slice_for_aabb` didn't correctly handle the special case of a ray parallel to sides of the AABB, which involves infinities.

- [x] Fix distance iterator
- [x] Deduplicates code of that iterator.
- [x] Breaking change in the signature of `Ray::intersection_slice_for_aabb` (`Option` is more idiomatic than `-1.0`)
- [x] Add a test case, which fails without the fix
- [x] Ensure that `Ray::intersection_slice_for_aabb` is still correct by either:
  - [x] testing that it still outputs the correct distance
image
  - [ ] ~~making it private and changing the spec, so that only relative distance matters~~
- [x] Assert regular traversal = distance traversal in fuzzer (now all 5 traversal types are consistent :smile:)

Fixes #119

## Future work
- Write the Ray-AABB tmin/tmax calculation only once (internally using SIMD if enabled), not 3 times

### DIFF
--- a/src/ray/ray_impl.rs
+++ b/src/ray/ray_impl.rs
@@ -2,7 +2,7 @@
 //! for axis aligned bounding boxes and triangles.
 
 use crate::aabb::IntersectsAabb;
-use crate::utils::{fast_max, fast_min};
+use crate::utils::fast_max;
 use crate::{aabb::Aabb, bounding_hierarchy::BHValue};
 use nalgebra::{
     ClosedAddAssign, ClosedMulAssign, ClosedSubAssign, ComplexField, Point, SVector, SimdPartialOrd,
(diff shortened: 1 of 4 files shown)


src/ray/ray_impl.rs
```diff
@@ -113,31 +113,26 @@ impl<T: BHValue, const D: usize> Ray<T, D> {
     /// where the first number is the distance from [`Ray`] to the nearest intersection point
     /// and the second number is the distance from [`Ray`] to the farthest intersection point
     ///
-    /// If there are no intersections, it returns negative one for both distances
-    pub fn intersection_slice_for_aabb(&self, aabb: &Aabb<T, D>) -> (T, T)
+    /// If there are no intersections, it returns `None`.
+    pub fn intersection_slice_for_aabb(&self, aabb: &Aabb<T, D>) -> Option<(T, T)>
```
svenstaro, line 117: Hm, this is API-breaking so please document the migration path in the CHANGELOG.

src/ray/ray_impl.rs
```diff
@@ -113,31 +113,26 @@ impl<T: BHValue, const D: usize> Ray<T, D> {
...
         // no intersection
-        if entry_distance > exit_distance || exit_distance < T::zero() {
-            return (T::from_f32(-1.0).unwrap(), T::from_f32(-1.0).unwrap());
+        if tmin > tmax || tmax < T::zero() {
+            return None;
         }
 
-        (fast_max(entry_distance, T::zero()), exit_distance)
+        Some((fast_max(tmin, T::zero()), tmax))
     }
```
svenstaro, line 136: I'll be honest, I didn't check the math but I trust you since the other tests still work. :P

svenstaro (changes requested):
Great change, trading library code for testing code and still coming in with fewer total lines!

I think the API change makes sense though it does need to be documented.

src/ray/ray_impl.rs
finnbear: Don't need to trust the math, since I added a test that the distances match those measured in blender for entry and exit. They were an exact match, modulo precision issues. As the comment states, the math is the same as ray-AABB intersection, modified to return the distance rather than comparing it.

src/ray/ray_impl.rs
finnbear: Sure! I added a changelog entry that also claimed the distance traversal was "correct." Then I realized it [might not be](https://github.com/svenstaro/bvh/issues/134). So now the wording is more precise.

svenstaro: Do you wanna spend some more time on that before we merge this?

PR #135, opened 2025-01-10
Docs #134 - acknowledge best-effort nature of dist traversal

### LINKED ISSUE #134: Investigate whether distance traversal iterator is correct
It seems to me that distance-traversing the following valid BVH would produce results in an incorrect order, because the left child AABB is hit first but contains the furthest object.

image

## TODO
- [x] Find/fuzz to find a concrete example
- [x] Fix the behavior or document it as "best effort"

finnbear:
The iterator is indeed capable of returning things in a slightly-incorrect order. I don't see an efficient way to fix it, and it may not matter for users, so this PR documents the limitation as-is.

Fixes #134

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fix some assertions [#129](https://github.com/svenstaro/bvh/pull/129) (thanks @finnbear)
 - Fix traversal in case of single-node BVH [#130](https://github.com/svenstaro/bvh/pull/130) (thanks @finnbear)
 - Fix intersection between rays and AABBs that lack depth. [#131](https://github.com/svenstaro/bvh/pull/131) (thanks @finnbear)
+- Document the limitations of distance traversal best-effort ordering. [#135](https://github.com/svenstaro/bvh/pull/135) (thanks @finnbear)
 
 ## 0.10.0 - 2024-07-06
 - Don't panic when traversing empty BVH [#106](https://github.com/svenstaro/bvh/pull/106) (thanks @finnbear)
--- a/src/bvh/bvh_impl.rs
+++ b/src/bvh/bvh_impl.rs
@@ -133,6 +133,9 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
     /// Returns a subset of [`shape`], in which the [`Aabb`]s of the elements were hit by [`Ray`].
     /// Return in order from nearest to farthest for ray.
     ///
+    /// This is a best-effort function that orders interior parent nodes before ordering child
+    /// nodes, so the output is not necessarily perfectly sorted.
+    ///
     /// [`Bvh`]: struct.Bvh.html
     /// [`Aabb`]: ../aabb/struct.AABB.html
     ///
@@ -147,6 +150,9 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
     /// Creates a [`DistanceTraverseIterator`] to traverse the [`Bvh`].
     /// Returns a subset of [`Shape`], in which the [`Aabb`]s of the elements were hit by [`Ray`].
     /// Return in order from farthest to nearest for ray.
+    ///
+    /// This is a best-effort function that orders interior parent nodes before ordering child
+    /// nodes, so the output is not necessarily perfectly sorted.
     ///
     /// [`Bvh`]: struct.Bvh.html
     /// [`Aabb`]: ../aabb/struct.AABB.html
--- a/src/bvh/distance_traverse.rs
+++ b/src/bvh/distance_traverse.rs
@@ -11,7 +11,10 @@ enum RestChild {
 }
 
 /// Iterator to traverse a [`Bvh`] in order from nearest [`Aabb`] to farthest for [`Ray`],
-/// without memory allocations
+/// or vice versa, without memory allocations.
+///
+/// This is a best-effort iterator that orders interior parent nodes before ordering child
+/// nodes, so the output is not necessarily perfectly sorted.
 pub struct DistanceTraverseIterator<
     'bvh,
     'shape,


svenstaro:
What would be the not-so-efficient way you could see to fix this?

finnbear:
To fix an arbitrarily incorrect order, the iterator would have to collect all the results, sort them, and then iterate the sorted collection.

In this case, the order is not *arbitrarily* incorrect. It may only be incorrect when there are non-`None` `RestChild`'s above in the stack. It might be possible to:
- keep track of whether any item in the traversal stack has a non-`None` `RestChild`
- if so, collect results into a `Vec` without outputting anything
- if not (or we reach the end), sort and output the `Vec` and then start outputting items directly

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fix some assertions [#129](https://github.com/svenstaro/bvh/pull/129) (thanks @finnbear)
 - Fix traversal in case of single-node BVH [#130](https://github.com/svenstaro/bvh/pull/130) (thanks @finnbear)
 - Fix intersection between rays and AABBs that lack depth. [#131](https://github.com/svenstaro/bvh/pull/131) (thanks @finnbear)
+- Document the limitations of distance traversal best-effort ordering. [#135](https://github.com/svenstaro/bvh/pull/135) (thanks @finnbear)
```
svenstaro, line 23: Well, it's not really a code change so I don't think we need a CHANGELOG entry for this.

svenstaro (commented):
So do you think we should merge this and have a `speedy_traverse()` and a `correct_traverse()` eventually or how should we play this? We can merge this for now and then see what we'll do?

CHANGELOG.md
finnbear: Hmm, I don't feel strongly about this, but it does seem worth noting. The function changed from claiming to produce ordered results (while having a bug) to claiming to produce not-necessarily-ordered results (while not having a bug). Even though the source code didn't changed, any user of the function should be put on notice to ensure their use-case doesn't depend on a property that doesn't hold.

Also, mentioning docs in changelogs is [not unprecedented](https://github.com/mrdoob/three.js/releases).

finnbear:
> We can merge this for now and then see what we'll do?

Sounds good to me. I don't have an opinion on next steps since I don't currently use this API.

Maybe we should ask @dashedman (author of https://github.com/svenstaro/bvh/pull/98) about whether accuracy or speed matters more :)

CHANGELOG.md
svenstaro: Works for me.

dashedman:
Oops

It seems that my algorithm works correctly only for those trees in which the children do not intersect with each other. This is an additional limitation for this algorithm.

Or if it is the order of the traverse through the tree that is important to you.

To solve the absolute pass in distance order, we can try to use a heap instead of a stack.

| Time complexity | Current algo (incorrect) | Algo with heap | Just sorting of all leafs |
| - | - | - | - |
| Get nearest first leaf     | O(log(n))    | O(log(n)^2)      | O(n*log(n)) |
| Get nearest first k leafs | O(k*log(n)) | O(k*log(n)^2) | O(n*log(n)) |
| Get all leafs in order     | O(n)           | O(n*log(n))      | O(n*log(n)) |

| Space complexity | Current algo (incorrect) | Algo with heap | Just sorting of all leafs |
| - | - | - | - |
| Get nearest first leaf     | O(log(n))  | O(log(n))    | O(log(n)) / O(n) (depends on sorting algo) |
| Get nearest first k leafs | O(log(n)) | O(k*log(n)) | O(log(n)) / O(n) (depends on sorting algo) |
| Get all leafs in order     | O(log(n))  | O(n)           | O(log(n)) / O(n) (depends on sorting algo) |

finnbear:
@dashedman thanks for your response!

> we can try to use a heap instead of a stack

I'm a bit new to BVH implementation. If you describe your proposal in a bit more detail, I will consider trying to implement it. Alternatively, you could submit a PR if you feel comfortable. No pressure, though.

dashedman:
If you give me test case code, it will be very helpfully

finnbear:
> If you give me test case code, it will be very helpfull

Sure, [here](https://github.com/svenstaro/bvh/issues/134#issuecomment-2584081007) is a ~minimized example.

dashedman:
image

can't compile lib

do you know what can it be?

finnbear:
@dashedman sorry about this, looks like `num-traits` [added the `clamp` function very recently](https://github.com/rust-num/num-traits/pull/305).

I recommend `cargo update --package num-traits`. I'll prepare a PR to fix this long-term.

dashedman:
Seems like your test passed

finnbear:
Hmm, it still fails for me. I ran `cargo update` to update all dependencies and it still failed. Not sure why this test isn't deterministic. For context, I'm using detail features and running on an i7-7700k with Ubuntu Linux 20.04 (not that it should matter).
```
---- ray::ray_impl::tests::test_incorrect_order stdout ----
thread 'ray::ray_impl::tests::test_incorrect_order' panicked at src/ray/ray_impl.rs:459:9:
assertion failed: bvh.nearest_traverse_iterator(&ray,
        &aabbs).is_sorted_by(|a, b|
        {
            let (a, _) = ray.intersection_slice_for_aabb(a).unwrap();
            let (b, _) = ray.intersection_slice_for_aabb(b).unwrap();
            a <= b
        })
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    ray::ray_impl::tests::test_incorrect_order
```

dashedman:
Of course, it passed on my local computer

dashedman:
Check please this PR

I'm gonna sleep

finnbear:
>  it passed...with new implementation

Oh!! congratulations! :heart:

PR #139, opened 2025-01-12
Fix distance traversing bug with heap

dashedman:
related to https://github.com/svenstaro/bvh/pull/135

### DIFF
--- a/src/bvh/bvh_impl.rs
+++ b/src/bvh/bvh_impl.rs
@@ -12,7 +12,10 @@ use crate::utils::joint_aabb_of_shapes;
 
 use std::mem::MaybeUninit;
 
-use super::{BvhNode, BvhNodeBuildArgs, DistanceTraverseIterator, ShapeIndex, Shapes};
+use super::{
+    BvhNode, BvhNodeBuildArgs, ChildDistanceTraverseIterator, DistanceTraverseIterator, ShapeIndex,
+    Shapes,
+};
 
 /// The [`Bvh`] data structure. Contains the list of [`BvhNode`]s.
 ///
@@ -133,8 +136,7 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
     /// Returns a subset of [`shape`], in which the [`Aabb`]s of the elements were hit by [`Ray`].
     /// Return in order from nearest to farthest for ray.
     ///
-    /// This is a best-effort function that orders interior parent nodes before ordering child
-    /// nodes, so the output is not necessarily perfectly sorted.
+    /// Time complexity: for first `O(log(n))`, for all `O(n*log(n))`
     ///
     /// [`Bvh`]: struct.Bvh.html
     /// [`Aabb`]: ../aabb/struct.AABB.html
@@ -151,8 +153,7 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
     /// Returns a subset of [`Shape`], in which the [`Aabb`]s of the elements were hit by [`Ray`].
     /// Return in order from farthest to nearest for ray.
     ///
-    /// This is a best-effort function that orders interior parent nodes before ordering child
-    /// nodes, so the output is not necessarily perfectly sorted.
+    /// Time complexity: for first `O(log(n))`, for all `O(n*log(n))`.
     ///
     /// [`Bvh`]: struct.Bvh.html
     /// [`Aabb`]: ../aabb/struct.AABB.html
@@ -163,6 +164,50 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
         shapes: &'shape [Shape],
     ) -> DistanceTraverseIterator<'bvh, 'shape, T, D, Shape, false> {
         DistanceTraverseIterator::new(self, ray, shapes)
+    }
+
+    /// Creates a [`ChildDistanceTraverseIterator`] to traverse the [`Bvh`].
+    /// Returns a subset of [`shape`], in which the [`Aabb`]s of the elements were hit by [`Ray`].
+    /// Return in order from nearest to farthest for ray.
+    ///
+    /// This is a best-effort function that orders interior parent nodes before ordering child
+    /// nodes, so the output is not necessarily perfectly sorted.
+    ///
+    /// So the output is not necessarily perfectly sorted if the children of any node overlap.
+    ///
+    /// Time complexity: for first `O(log(n))`, for all `O(n)`.
+    ///
+    /// [`Bvh`]: struct.Bvh.html
+    /// [`Aabb`]: ../aabb/struct.AABB.html
+    ///
+    pub fn nearest_child_traverse_iterator<'bvh, 'shape, Shape: Bounded<T, D>>(
+        &'bvh self,
+        ray: &'bvh Ray<T, D>,
+        shapes: &'shape [Shape],
+    ) -> ChildDistanceTraverseIterator<'bvh, 'shape, T, D, Shape, true> {
+        ChildDistanceTraverseIterator::new(self, ray, shapes)
+    }
+
+    /// Creates a [`ChildDistanceTraverseIterator`] to traverse the [`Bvh`].
+    /// Returns a subset of [`Shape`], in which the [`Aabb`]s of the elements were hit by [`Ray`].
+    /// Return in order from farthest to nearest for ray.
+    ///
+    /// This is a best-effort function that orders interior parent nodes before ordering child
+    /// nodes, so the output is not necessarily perfectly sorted.
+    ///
+    /// So the output is not necessarily perfectly sorted if the children of any node overlap.
+    ///
+    /// Time complexity: for first `O(log(n))`, for all `O(n)`.
+    ///
+    /// [`Bvh`]: struct.Bvh.html
+    /// [`Aabb`]: ../aabb/struct.AABB.html
+    ///
+    pub fn farthest_child_traverse_iterator<'bvh, 'shape, Shape: Bounded<T, D>>(
+        &'bvh self,
+        ray: &'bvh Ray<T, D>,
+        shapes: &'shape [Shape],
+    ) -> ChildDistanceTraverseIterator<'bvh, 'shape, T, D, Shape, false> {
+        ChildDistanceTraverseIterator::new(self, ray, shapes)
     }
 
     /// Prints the [`Bvh`] in a tree-like visualization.
--- a/src/bvh/child_distance_traverse.rs
+++ b/src/bvh/child_distance_traverse.rs
@@ -1,0 +1,336 @@
+use crate::aabb::Bounded;
+use crate::bounding_hierarchy::BHValue;
+use crate::bvh::{iter_initially_has_node, Bvh, BvhNode};
+use crate::ray::Ray;
+
+#[derive(Debug, Clone, Copy)]
+enum RestChild {
+    Left,
+    Right,
+    None,
+}
+
+/// Iterator to traverse a [`Bvh`] in order from nearest [`Aabb`] to farthest for [`Ray`],
+/// or vice versa, without memory allocations.
+///
+/// This is a best-effort iterator that orders interior parent nodes before ordering child
+/// nodes, so the output is not necessarily perfectly sorted.
+pub struct ChildDistanceTraverseIterator<
+    'bvh,
+    'shape,
+    T: BHValue,
+    const D: usize,
+    Shape: Bounded<T, D>,
+    const ASCENDING: bool,
+> {
+    /// Reference to the Bvh to traverse
+    bvh: &'bvh Bvh<T, D>,
+    /// Reference to the input ray
+    ray: &'bvh Ray<T, D>,
+    /// Reference to the input shapes array
+    shapes: &'shape [Shape],
+    /// Traversal stack. 4 billion items seems enough?
+    stack: [(usize, RestChild); 32],
+    /// Position of the iterator in bvh.nodes
+    node_index: usize,
+    /// Size of the traversal stack
+    stack_size: usize,
+    /// Whether or not we have a valid node (or leaf)
+    has_node: bool,
+}
+
+impl<'bvh, 'shape, T, const D: usize, Shape: Bounded<T, D>, const ASCENDING: bool>
+    ChildDistanceTraverseIterator<'bvh, 'shape, T, D, Shape, ASCENDING>
+where
+    T: BHValue,
+{
+    /// Creates a new [`DistanceTraverseIterator`]
+    pub fn new(bvh: &'bvh Bvh<T, D>, ray: &'bvh Ray<T, D>, shapes: &'shape [Shape]) -> Self {
+        ChildDistanceTraverseIterator {
+            bvh,
+            ray,
+            shapes,
+            stack: [(0, RestChild::None); 32],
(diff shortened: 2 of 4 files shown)


finnbear:
Very cool!

I modified the fuzzer to assert correctness and, in `Mode::Grid` (well-behaved AABB and Ray coordinates), `nearest_traverse_iterator` passes! (no errors for ~5m on my machine).

In `Mode::Chaos`, `nearest_traverse_iterator`'s order wasn't always correct but, in this mode, rounding errors could easily explain the issue.

`farthest_traverse_iterator` fails regardless of the fuzzer `Mode`, especially when the ray starts within one of the AABB's. I'm still trying to figure out why ;)

dashedman:
What is the fuzzer?

And it fails `farthest_traverse_iterator` with new implementation?

dashedman:
изображение

It's depends on what we want from farthest algo
Farthest by entry point or exit points? 

Now it's working from entry points

finnbear:
> What is the fuzzer?

See `/fuzz/fuzz_targets/fuzz.rs` (comment at beginning of file and end of README.md).

Here are my modifications to the fuzzer: https://gist.github.com/finnbear/4f4323814ff62c24cc96c8fab4c24d13

1. install `cargo-fuzz`
2. `cargo fuzz run fuzz`

dashedman:
I think we should change algo for exit point

dashedman:
please check

svenstaro:
I probably won't have time to give this a full review very soon so it would be amazing if @finnbear and perhaps even @Azkellas could look at this if you have some time. :)

finnbear:
Cool; I started fuzzing this, by switching to `ray.intersection_slice_for_aabb(&shape.aabb()).unwrap().1` on the farthest->nearest direction.

<details>
<summary>First thing that happened was I observed a NaN distance, which I'll file away here for later investigation. I don't see how this could possibly be due to your changes.</summary>

```
Workload {
            shapes: [
                Aabb {
                    min: [
                        5.693442e-39,
                        -5000.0,
                        5000.0,
                    ],
                    max: [
                        5.693442e-39,
                        -5000.0,
                        5000.0,
                    ],
                },
            ],
            ray: Ray {
                origin: [
                    -5000.0,
                    -5000.0,
                    5000.0,
                ],
                direction: [
                    [
                        0.70710677,
                        0.70710677,
                        0.0,
                    ],
                ],
                inv_direction: [
                    [
                        1.4142135,
                        1.4142135,
                        inf,
                    ],
                ],
            },
            point: [
                1.1859499e-23,
                7.9149854e-24,
                5000.0,
            ],
            aabb: Aabb {
                min: [
                    195.26274,
                    4.53e-43,
                    0.0,
                ],
                max: [
                    195.26274,
                    4.53e-43,
                    0.0,
                ],
            },
            ball: Ball {
                center: [
                    195.26274,
                    195.26274,
                    195.26274,
                ],
                radius: 195.26274,
            },
            mutations: [
                Add(
                    Aabb {
                        min: [
                            0.0,
                            0.0,
                            0.0,
                        ],
                        max: [
                            9.1834e-41,
                            1.0,
                            1.0,
                        ],
                    },
                ),
            ],
        }
```
</details>

I re-enabled `Mode::Grid` to avoid this. At that point, both directions of traversals output correct orderings. I cannot currently explain why switching from nearest to farthest point made errors go away.

My tentative conclusion is that 1) this PR implements a correct algorithm 2) the ray slice calculation needs some investigation into NaN.

Due to the NaN issue, I cannot say whether this PR's algorithm is susceptible to rounding errors.

dashedman:
Most probably NaN is the result that some coord of ray direction is zero

dashedman:
https://github.com/svenstaro/bvh/blob/204a1c984ad99457263bb7e3ae6c963301d31ef7/src/ray/ray_impl.rs#L77

you can try add here `if x == T::zero() { T:inf() } else { T:zero() / x}`

src/bvh/bvh_impl.rs
```diff
@@ -165,6 +166,50 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
...
+    /// nodes, so the output is not necessarily perfectly sorted.
+    ///
+    /// Useful if childs don't intersect each other.
+    ///
+    /// Time complexity: for first O(log(n)), for all O(n).
+    ///
+    /// [`Bvh`]: struct.Bvh.html
+    /// [`Aabb`]: ../aabb/struct.AABB.html
+    ///
+    pub fn nearest_by_child_traverse_iterator<'bvh, 'shape, Shape: Bounded<T, D>>(
```
finnbear, line 183: ## *Tactical* review
 - Please remove the word `by_` from the name, for consistency with `farthest_child_traverse_iterator`.
 - I find "Useful if childs don't intersect each other." to be misleading, because it may be useful even if children intersect (if an inexact result is permissible). I think it would be better to alter the previous paragraph to end with "so the output is not necessarily perfectly sorted **if the children of any node overlap**.
 - `DistanceTraverseIterator` in the docs is a typo and should be edited to match the final name of the iterator.

This comment applies to `farthest_child_traverse_iterator` and also implies renaming `DistanceByChildTraverseIterator`.

## *Strategic* Review
Alternatively, the old iterator could be removed. I'm not personally familiar with the use-cases of approximate traversal but I'm pretty sure a considerable fraction of normal BVH's will have intersecting children at some level, and I'm not aware of any workaround to prevent it.

If the expectation is that users engineer their input to the BVH building algorithm to never produce intersecting children, I don' think that expectation is a realistic one.

## Conclusion
If you don't personally have a use-case for the faster/incorrect version and cannot think of a clear use-case that anyone else would have, I recommend removing it from this PR. It can always be resurrected from `git` in a future PR (e.g. if someone complains).

src/bvh/distance_traverse.rs
```diff
@@ -1,13 +1,36 @@
...
+struct DistNodePair<T: PartialOrd> {
+    dist: T,
+    node_index: usize,
+}
+
+impl<T: PartialOrd> Eq for DistNodePair<T> {}
+
+impl<T: PartialOrd> PartialEq<Self> for DistNodePair<T> {
+    fn eq(&self, other: &Self) -> bool {
+        self.dist.eq(&other.dist)
```
finnbear, line 19: This allows silently returning `false` in the presence of `NaN`, which violates reflexivity, which violates the spec of `Eq`, which could lead to bad outcomes when used in `std::collections` collection.

A better implementation would be `self.dist.cmp(&other.dist) == Ordering::Equal` unless `clippy` complains, in which case try `self.dist.partial_cmp(&other.dist).unwrap() == Ordering::Equal`.

I'd also add a comment regarding that `NaN` issue.

src/bvh/distance_traverse.rs
```diff
@@ -1,13 +1,36 @@
...
+impl<T: PartialOrd> PartialOrd<Self> for DistNodePair<T> {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        self.dist.partial_cmp(&other.dist)
+    }
+}
+
+impl<T: PartialOrd> Ord for DistNodePair<T> {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.partial_cmp(other).unwrap()
+    }
```
finnbear, line 33: Let's avoid [the lint](https://rust-lang.github.io/rust-clippy/rust-1.84.0/index.html#non_canonical_partial_ord_impl) entirely by putting `self.dist.partial_cmp(&other.dist).unwrap()` in `Ord` and `Some(self.cmp(other))` in `PartialOrd`.

src/bvh/distance_traverse.rs
```diff
@@ -46,133 +63,75 @@ where
...
+                let left_dist_opt = self.ray.intersection_slice_for_aabb(child_l_aabb);
+                let right_dist_opt = self.ray.intersection_slice_for_aabb(child_r_aabb);
+
+                if let Some((left_dist_min, left_dist_max)) = left_dist_opt {
+                    // left child has intersection with ray
+                    let left_dist = if ASCENDING {
+                        left_dist_min
+                    } else {
+                        left_dist_max
+                    };
```
finnbear, line 101: I think this deserves a comment *somewhere*. Perhaps the public documentation should mention that farthest-traversal considers the distance to the exit point of the ray. Perhaps that should also be mentioned here.

src/bvh/distance_traverse.rs
```diff
@@ -46,133 +63,75 @@ where
...
-                }
-            },
-            BvhNode::Leaf { .. } => {
-                self.has_node = false;
-            }
-        }
+    fn add_to_heap(&mut self, dist_to_node: T, node_index: usize) {
+        // cause we use max-heap, it store max value on the top
+        if ASCENDING {
+            // we need the smallest distance, so we negotiate value
```
finnbear, line 123: so we **negate** the value

src/bvh/distance_traverse.rs
```diff
@@ -333,4 +280,45 @@ mod tests {
...
+            unimplemented!();
+        }
+
+        fn set_bh_node_index(&mut self, _: usize) {
+            // No-op.
+        }
+    }
+
+    #[test]
+    fn test_incorrect_order() {
```
finnbear, line 295: Now that the order is correct, this could say `test_overlapping_child_order`.

finnbear (changes requested):
Thanks for these changes! :rocket:

I've left some feedback that I hope you can address (if not, that's fine; I will personally drive the PR to completion rather than letting it become too stale).

I would like to see the fuzzer get updated to assert correct order. Feel free to add that, or I can do it in a later PR.

As a final note, to the extent you feel comfortable doing so, please `git rebase` your changes into meaningful commits. You can err on the side of squashing commits, especially those that "fix" prior commits. If you end up with just one commit, that's totally fine.

src/bvh/bvh_impl.rs
dashedman: I develop this algo for my task, and in my task was BVH with no intersecting children. 

@svenstaro what do you think? Do we need one more algo for the specific BVH case (when the children of any node overlap)?

dashedman:
Done all except eq trait and removing old algo

src/bvh/distance_traverse.rs
dashedman: image

Can I do this? Clippy doesn't complain

src/bvh/bvh_impl.rs
```diff
@@ -133,8 +133,7 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
     /// Returns a subset of [`shape`], in which the [`Aabb`]s of the elements were hit by [`Ray`].
     /// Return in order from nearest to farthest for ray.
     ///
-    /// This is a best-effort function that orders interior parent nodes before ordering child
-    /// nodes, so the output is not necessarily perfectly sorted.
+    /// Time complexity: for first O(log(n)), for all O(n*log(n))
```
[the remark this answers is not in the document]
dashedman: done

src/bvh/distance_traverse.rs
dashedman: done

src/bvh/bvh_impl.rs
dashedman: @finnbear I'm inclined to leave the code. It doesn't ask for food

finnbear: Lol, works for me!

src/bvh/bvh_impl.rs
```diff
@@ -165,6 +166,50 @@ impl<T: BHValue, const D: usize> Bvh<T, D> {
...
     }
 
+    /// Creates a [`ChildDistanceTraverseIterator`] to traverse the [`Bvh`].
+    /// Returns a subset of [`shape`], in which the [`Aabb`]s of the elements were hit by [`Ray`].
+    /// Return in order from nearest to farthest for ray.
+    ///
+    /// This is a best-effort function that orders interior parent nodes before ordering child
+    /// nodes, so the output is not necessarily perfectly sorted.
+    ///
+    /// So the output is not necessarily perfectly sorted if the children of any node overlap.
```
finnbear, line 176:
```
    /// This is a best-effort function that orders interior parent nodes before ordering child
    /// nodes, so the output is not necessarily perfectly sorted.
    ///
    /// So the output is not necessarily perfectly sorted if the children of any node overlap.
```
should have been
```
    /// This is a best-effort function that orders interior parent nodes before ordering child
    /// nodes, so the output is not necessarily perfectly sorted if the children of any node
    /// overlap.
```

src/bvh/child_distance_traverse.rs
```diff
@@ -0,0 +1,336 @@
...
+    fn traverse_distance_and_verify_order(
+        ray_origin: TPoint3,
+        ray_direction: TVector3,
+        all_shapes: &[UnitBox],
+        bvh: &TBvh3,
+        expected_shapes: &HashSet<i32>,
+    ) {
+        let ray = Ray::new(ray_origin, ray_direction);
+        let near_it = bvh.nearest_traverse_iterator(&ray, all_shapes);
+        let far_it = bvh.farthest_traverse_iterator(&ray, all_shapes);
```
finnbear, line 250: Should change old tests to point to the old iterator under the new name, not the new iterator under the old name.

src/bvh/child_distance_traverse.rs
```diff
@@ -0,0 +1,336 @@
...
+        let (shapes, bvh) = build_empty_bvh();
+
+        // Define an arbitrary ray.
+        let origin = TPoint3::new(0.0, 0.0, 0.0);
+        let direction = TVector3::new(1.0, 0.0, 0.0);
+        let ray = Ray::new(origin, direction);
+
+        // Ensure distance traversal doesn't panic.
+        assert_eq!(bvh.nearest_traverse_iterator(&ray, &shapes).count(), 0);
+        assert_eq!(bvh.farthest_traverse_iterator(&ray, &shapes).count(), 0);
```
finnbear, line 334: Same thing here, too :)

finnbear:
To fix the fuzzer, please copy my changes from #142 into this PR:
image

Or, we could merge #142 and you could rebase onto master. Idk what is easier.

finnbear:
done!

finnbear (approved):
LGTM

## TODO
- [ ] rebase into meaningful commits, squashing any commits that "fix" a previous commit. you may end up with only 1-2 large commits.

## Future work (may come after this PR)
- [ ] Assert distance ordering in fuzzer
- [ ] Changelog edits

dashedman:
done

finnbear:
Thanks; could you reword your commit message to be something like "Fix order of distance traversal, preserving old algorithm under a different name" rather than including the word "squash" ?

dashedman:
I don't know what in github, but in gitlab there is option for auto squash when PR is merged

finnbear:
Yeah, there is a Squash and Merge option in GitHub. However, when reviewing my PR's, the original maintainer has preferred manual rebase (possibly into more than one meaningful commit) followed by normal merge.

finnbear (approved):
As far as I'm concerned, this is ready to be merged :)

I'll give @svenstaro a chance to quickly review everything, just in case.

svenstaro:
@finnbear If you think it looks good then I think it's good to go.

finnbear:
Working on a changelog/fuzz PR now :)

Issue #140, opened 2025-01-13
Investigate case of NaN/infinity returned by `Ray::intersection_slice_for_aabb`


finnbear:
While reviewing #139, I noticed the distance returned by `Ray::intersection_slice_for_aabb` was `NaN` despite the input being finite.

- [x] Create a test case that reproduces the issue
- [x] Prevent this from happening

Here was the context in which it occurred:
```
Workload {
            shapes: [
                Aabb {
                    min: [
                        5.693442e-39,
                        -5000.0,
                        5000.0,
                    ],
                    max: [
                        5.693442e-39,
                        -5000.0,
                        5000.0,
                    ],
                },
            ],
            ray: Ray {
                origin: [
                    -5000.0,
                    -5000.0,
                    5000.0,
                ],
                direction: [
                    [
                        0.70710677,
                        0.70710677,
                        0.0,
                    ],
                ],
                inv_direction: [
                    [
                        1.4142135,
                        1.4142135,
                        inf,
                    ],
                ],
            },
        }
```

finnbear:
Investigated and found the cause.

The issue occurs when a ray origin lies in the plane of an AABB face while also being tangent to to that face.

This also affects normal intersection. For example, this test should succeed but doesn't.
```rust
    #[test]
    fn test_ray_origin_in_plane_of_and_tangent_to_aabb_face() {
        let aabb = TAabb3::with_bounds(TPoint3::new(-0.5, -0.5, -0.5), TPoint3::new(0.5, 0.5, 0.5));
        let ray = TRay3::new(
            TPoint3::new(0.0, 0.0, 0.5),
            TVector3::new(0.0, -1.0, 0.0),
        );
        assert!(ray.intersects_aabb(&aabb));
    }
```

Arguably, this isn't much of a problem for normal intersection, because the result is that rays may miss if they were very close to missing anyway.

For `intersection_slice_for_aabb`, returning NaN seems like a bigger problem. Next step is to find a solution that doesn't hurt performance (too much).

finnbear:
It appears that my fix only works for one axis, because of how `fast_min` and `fast_max` are order-dependent on NaN's. Will need a new strategy.

PR #143, opened 2025-01-14
Changelog edit and fuzz for #139

finnbear:
Followup to #139

- [x] Changelog
- [x] Fuzz (make sure ordering of new algorithm is correct)

Noticed #140 (infinity this time, not NaN), which needs a new fix, but that's orthogonal (only affects near-miss rays, so I simply restrict the new assertions to `Mode::Grid`)

<sup>I intend to wait ~24h before merging, just in case there are any comments/suggestions/objections.</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,8 @@ jobs:
 
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: nightly
+          # use a known good version (e.g. on Jan 14 2025, a faulty `rustc` nightly panicked)
+          toolchain: nightly-2024-01-13
           components: rust-src
 
       - name: Install cargo-fuzz
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   allowing points, AABB's, and circles/spheres to be tested, too. Most use-cases involving `Ray` 
   will continue to compile as-is. If you previously wrote `BvhTraverseIterator<T, D, S>`, you'll
   need to change it to `BvhTraverseIterator<T, D, Ray, S>`. [#128](https://github.com/svenstaro/bvh/pull/128) (thanks @finnbear)
-- **Breaking change:** Distance-traversal no longer outputs non-intersected shapes, but note that
-  `Ray::intersection_slice_for_aabb` now returns `None` instead of `(-1.0, -1.0)` in the case of no 
+- **Breaking change:** `Ray::intersection_slice_for_aabb` now returns `None` instead of `(-1.0, -1.0)` in the case of no 
   intersection, and `Some((entry, exit))` in the case of intersection. [#133](https://github.com/svenstaro/bvh/pull/133) [#142](https://github.com/svenstaro/bvh/pull/142) (thanks @finnbear)
+- `Bvh::nearest_traverse_iterator` and `Bvh::farthest_traverse_iterator` now output correctly ordered results when the children
+  of an internal node overlap, resulting in them taking more time and requiring heap allocation.
+  The new iterators `Bvh::nearest_child_traverse_iterator` and `Bvh::farthest_child_traverse_iterator` use the old algorithm. [#133](https://github.com/svenstaro/bvh/pull/139) (thanks @dashedman)
 - Fix panic on empty `DistanceTraverseIterator` [#117](https://github.com/svenstaro/bvh/pull/117) (thanks @finnbear)
 - Fix center() for very large AABBs [#118](https://github.com/svenstaro/bvh/pull/118) (thanks @finnbear)
 - Fix more cases where an empty BVH would panic [#116](https://github.com/svenstaro/bvh/pull/116) (thanks @finnbear)
--- a/fuzz/fuzz_targets/fuzz.rs
+++ b/fuzz/fuzz_targets/fuzz.rs
@@ -16,8 +16,10 @@
 
 use std::cmp::Ordering;
 use std::collections::HashSet;
+use std::env;
 use std::fmt::{self, Debug, Formatter};
 use std::hash::{Hash, Hasher};
+use std::str::FromStr;
 
 use arbitrary::Arbitrary;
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -10,9 +10,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   allowing points, AABB's, and circles/spheres to be tested, too. Most use-cases involving `Ray` 
   will continue to compile as-is. If you previously wrote `BvhTraverseIterator<T, D, S>`, you'll
   need to change it to `BvhTraverseIterator<T, D, Ray, S>`. [#128](https://github.com/svenstaro/bvh/pull/128) (thanks @finnbear)
-- **Breaking change:** Distance-traversal no longer outputs non-intersected shapes, but note that
-  `Ray::intersection_slice_for_aabb` now returns `None` instead of `(-1.0, -1.0)` in the case of no 
+- **Breaking change:** `Ray::intersection_slice_for_aabb` now returns `None` instead of `(-1.0, -1.0)` in the case of no 
   intersection, and `Some((entry, exit))` in the case of intersection. [#133](https://github.com/svenstaro/bvh/pull/133) [#142](https://github.com/svenstaro/bvh/pull/142) (thanks @finnbear)
+- `Bvh::{nearest, farthest}_traverse_iterator` now output correctly ordered results when the children
+  of an internal node overlap, resulting in them taking more time and requiring heap allocation.
+  The new `Bvh::{nearest, farthest}_child_traverse_iterator` use the old algorithm. [#133](https://github.com/svenstaro/bvh/pull/139) (thanks @dashedman)
```
svenstaro, line 17: Make this easier to search for:
```suggestion
- `Bvh::nearest_traverse_iterator` and `Bvh::farthest_traverse_iterator` now output correctly ordered results when the children
  of an internal node overlap, resulting in them taking more time and requiring heap allocation.
  The new iterators `Bvh::nearest_child_traverse_iterator` and `Bvh::farthest_child_traverse_iterator` use the old algorithm. [#133](https://github.com/svenstaro/bvh/pull/139) (thanks @dashedman)
```

fuzz/fuzz_targets/fuzz.rs
```diff
@@ -334,6 +334,13 @@ impl<const D: usize> Workload<D> {
             assert!(!intersects);
         }
 
+        // HINT: uncomment this temporarily to test with small BVH's.
+        /*
+        if !self.mutations.is_empty() || self.shapes.len() > 5 {
+            return;
+        }
+        */
```
svenstaro, line 342: Hm, that's a bit weird to keep it here. Can we find a solution where we'll benefit from this permanently and not just when it's commented in? You can also keep it this way if you think it's actually helpful like this.

fuzz/fuzz_targets/fuzz.rs
```diff
@@ -390,23 +397,55 @@ impl<const D: usize> Workload<D> {
...
+                })
+                .map(ByPtr)
+                .collect::<HashSet<_>>();
+
+            let nearest_child_traverse_iterator = bvh
+                .nearest_child_traverse_iterator(&ray, &self.shapes)
+                .map(ByPtr)
+                .collect::<HashSet<_>>();
+            let farthest_child_traverse_iterator = bvh
+                .farthest_child_traverse_iterator(&ray, &self.shapes)
```
svenstaro, line 432: This whole section is a bit sparse in terms of comments (but clearly this is not just boilerplate). Could you document the intent here?

svenstaro (changes requested):

fuzz/fuzz_targets/fuzz.rs
finnbear: The idea is that we want to detect bugs in large BVH's (or bugs from mutations). Once a bug is detected, it **may** be reproducible with a simpler BVH. I find myself adding this sort of thing while debugging almost every time.

I have an idea, though. We could run two instances of the fuzzer in parallel, with one version checking all BVH's and one checking small ones. Then, the output of the small BVH fuzzer may show simple issues, without the risk of missing issues that only affect large BVH.

fuzz/fuzz_targets/fuzz.rs
finnbear: done :)

.github/workflows/ci.yml
```diff
@@ -54,7 +54,7 @@ jobs:
 
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: nightly
```
finnbear, line 57: Latest nightly refuses to compile the code. This is a temporary workaround.
image

.github/workflows/ci.yml
```diff
@@ -54,7 +54,8 @@ jobs:
...
+          # use a known good version (e.g. on Jan 14 2025, a faulty `rustc` nightly panicked)
+          toolchain: nightly-2024-01-13
```
finnbear, line 58: Latest nightly refuses to compile the code. This can be a temporary workaround, but may also make the CI more robust.
image

svenstaro: I'm not sure. I think it's good for the ecosystem to always run the latest nightly. We will otherwise definitely forget to upgrade.

svenstaro (approved):
Works for me!

.github/workflows/ci.yml
finnbear: > I think it's good for the ecosystem to always run the latest nightly.

Yes, it is possible to sacrifice reliability on our end to increase the chance that we find (and report?) issues to upstream `rustc` for the benefit of the ecosystem.

Never upgrading is also a bad policy, as eventual developments in locally-installed `nightly` Rust will produce code that doesn't work on the CI's version.

I personally view the harms of `rustc`-roulette to be greater, as it creates [collateral damage](https://github.com/svenstaro/bvh/pull/108#issuecomment-2594038147) to unrelated PR's. Never upgrading only risks errors when updating the fuzzer itself.

A real solution might involve trying the latest and producing a warning while falling back to a known good version if it doesn't work, but it seems complex to implement in CI.

I'll merge, but I'll make a tracking issue to help avoid forgetting about this.

PR #149, opened 2025-01-26
Fix #140 part 2 - detect NaN when intersection ray with AABB

### LINKED ISSUE #140: Investigate case of NaN/infinity returned by `Ray::intersection_slice_for_aabb`

finnbear:
I got a bit distracted over the past week but, when I took a look this afternoon, I found a complete solution for #140.

The change is to consider all rays that are in the plane of an AABB face as non-intersecting with the AABB, by detecting NaN. Previously, a subset of those rays were intersecting, typically with NaN/infinity/finite-but-wrong intersection slices. It seems very difficult to assign a correct intersection slice to those rays.

A NaN results when a coordinate of the ray origin matches the corresponding coordinate of an AABB face AND the ray's direction has a value of 0 for that axis. This causes 0/0 = NaN, which is detectable.

## Testing
- [x] The included test fails without this change
- [x] The included fuzzer change fails without this change

## Benchmarking
I made some effort to optimize the code. Here is a quick benchmark:
### No SIMD
```
cargo bench --features bench -- bench_intersect_12k_triangles_bvh
before: 363.77ns
after: 382.76ns
regression: 5.2%
```
### SIMD
```
cargo bench --features bench,simd -- bench_intersect_12k_triangles_bvh
before: 294.96ns
after: 304.78ns
regression: 3.3%
```

Fixes #140 (again)

### DIFF
--- a/fuzz/fuzz_targets/fuzz.rs
+++ b/fuzz/fuzz_targets/fuzz.rs
@@ -381,13 +381,27 @@ impl<const D: usize> Workload<D> {
         let ray = self.ray.ray();
         let aabb = self.aabb.aabb();
 
-        // Make sure these functions agree and that the latter doesn't return NaN.
+        // Make sure these functions agree and that the latter doesn't return NaN/infinity.
         let intersects = ray.intersects_aabb(&aabb);
         if let Some((entry, exit)) = ray.intersection_slice_for_aabb(&aabb) {
-            assert!(intersects);
-            assert!(entry <= exit, "{entry} {exit}");
+            assert!(
+                intersects,
+                "intersection_slice_for_aabb returned Some for non-intersection"
+            );
+            assert!(entry >= 0.0, "entry ({entry}) must be non-negative");
+            assert!(
+                entry <= exit,
+                "entry ({entry}) must not exceed exit ({exit})"
+            );
+            assert!(
+                entry.is_finite() && exit.is_finite(),
+                "neither entry ({entry}) nor exit ({exit}) may be infinite or NaN for AABB's and rays with finite coordinates"
+            );
         } else {
-            assert!(!intersects);
+            assert!(
+                !intersects,
+                "intersection_slice_for_aabb returned None for intersection"
+            );
         }
 
         // If this environment variable is present, only test limited-size BVH's without mutations.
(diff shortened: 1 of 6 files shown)


fuzz/fuzz_targets/fuzz.rs
```diff
@@ -381,11 +381,14 @@ impl<const D: usize> Workload<D> {
...
         let intersects = ray.intersects_aabb(&aabb);
         if let Some((entry, exit)) = ray.intersection_slice_for_aabb(&aabb) {
             assert!(intersects);
-            assert!(entry <= exit, "{entry} {exit}");
+            assert!(
+                entry >= 0.0 && entry <= exit && entry.is_finite() && exit.is_finite(),
+                "{entry} {exit}"
+            );
         } else {
             assert!(!intersects);
```
svenstaro, line 393: Maybe, since you're here anyway, document via comment or preferably via second `assert!()` argument, what these asserts are even looking for.

src/utils.rs
```diff
@@ -114,3 +115,8 @@ pub(crate) fn joint_aabb_of_shapes<T: BHValue, const D: usize, Shape: BHShape<T,
     }
     (aabb, centroid)
 }
+
+#[inline(always)]
+pub(crate) fn has_nan<'a, T: Float + 'a>(iter: impl IntoIterator<Item = &'a T>) -> bool {
+    iter.into_iter().any(|f| f.is_nan())
+}
```
svenstaro, line 122: I know the other methods here aren't documented but let's start here. :)

svenstaro (approved):
Seems good apart from a little bit of missing docs. :)

svenstaro (approved):

svenstaro:
Do you also want to add a CHANGELOG entry?

svenstaro:
Excellent!

PR #152, opened 2025-01-27
Fancy benchmarks

### LINKED ISSUE #122: Rework benchmarks in README
https://github.com/svenstaro/bvh/issues/115#issuecomment-2572624609: We could have a pre-collapsed performance sections for people that really care about the numbers. However, we should probably get rid of the numbers in the README altogether. They are very old anyway and at this point likely misleading.

My thoughts: It would be nice to have a graph comparing performance with other BVH implementations, especially other Rust crates.

finnbear:
This is WIP; some improvements are necessary, and I'm not 100% sure this is a good idea.

fancy_benchmark_2
fancy_benchmarks_3

(aspirationally) Fixes #122

### DIFF
--- a/benches/Cargo.toml
+++ b/benches/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "benches"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+bvh = { path = ".." }
+clap = { version = "4.5.27", features = ["derive"] }
+nalgebra = "0.33.2"
+rand = "0.9.0"
--- a/benches/plot.py
+++ b/benches/plot.py
@@ -5,55 +5,71 @@ import math
 from subprocess import Popen, PIPE
 
 # number of benchmarks per power of 10
-resolution = 10
+resolution = 4
 # benchmark 1..=10^max rays/triangles
 max = 4
 
+all_benchmarks = [
+    "traverse",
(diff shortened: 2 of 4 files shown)


svenstaro:
If I'm reading this right, the question the graphs try to answer is "How much faster is using a BVH over bruteforcing for the same parameters". I'm not sure I'm parsing this right but I don't think the speedup would be maxing out at a factor of two since we should be looking at exponential gains over a bruteforce implementation.

finnbear:
Yeah, the speedup has no upper bound. It's currently capped at 2 because it made the color scheme simple, but that could be changed.

Capping at 2, in a clearer way, could still be useful since it shows when a BVH is faster (albeit not how much faster).

svenstaro:
I think the current way is a bit misleading. In general, I'm not sure how readable this three-dimensional graph can be made considering colors doesn't map so well to logarithmic scale (or at least it's not intuitive). How about a set of two-dimensional graphs that show a logarithmic speedup?

svenstaro:
Also as a general thing, I might be _cool_ to use [plotters](https://github.com/plotters-rs/plotters) here :crab:

finnbear:
I agree with all your feedback and have some ideas but, as I don't currently have the time to implement them, I'll close this for now. Might re-open and finish later.

PR #161, opened 2025-07-26
Remove extra deps and put log behind feature flag.

caibear:
On my computer running `cargo clean && cargo build --no-default-features`
Before: ```Finished `dev` profile [unoptimized + debuginfo] target(s) in 10.07s```
After: ```Finished `dev` profile [unoptimized + debuginfo] target(s) in 5.40s```
Which linearly follows the number of crates needed 39 -> 17.

The only API change is putting 2 log::info in bvh optimization behind a feature flag. These logging calls could also be removed if they aren't needed anymore.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,25 +16,28 @@ keywords = ["bvh", "bounding", "volume", "sah", "aabb"]
 license = "MIT"
 
 [dependencies]
-approx = "0.5"
-log = "0.4"
-serde = { optional = true, version = "1", features = ["derive"] }
-num = "0.4.3"
-nalgebra = { version = "0.33.0", features = ["default", "serde-serialize"] }
+log = { optional = true, version = "0.4" }
+nalgebra = { version = "0.33.0", default-features = false }
+num-traits = { version = "0.2.19", default-features = false, features = ["std"] }
 rayon = { optional = true, version = "1.8.1" }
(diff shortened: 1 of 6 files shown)


finnbear (approved):
Thank you!

List of changes:
- Make `log` (and two `log::info`'s inside optimization) optional
- Removing `nalgebra` `default` feature
- Condition `nalgebra` `serde-serialize` feature on `bvh` `serde` feature
- Replace `num` with default features with `num-traits` with `std` feature only
- Condition `wide` on `bvh` `simd` feature
- Move `approx` to dev-dependencies
- Use `nalgebra` again, with default features, in dev-dependencies
- Use `num` again, with default features, in dev-dependencies
- Alphabetize dependencies and dev-dependencies

This is a breaking change, but only because users cannot rely on importing `bvh` to enable `nalgebra`'s `serde-serialize` feature. I consider this acceptable breakage.

The `clippy` failure is unrelated.

finnbear:
> Perhaps a series of smaller MRs or commits would be easier to reason about in case this breaks something?

I opened #163 for this purpose. This PR can be edited to only contain the `nalgebra` stuff, possibly with an argument of why the changes are compatible (since it's the only non-trivial one).

svenstaro:
@caibear Do you wanna pick up work on this?

PR #164, opened 2025-07-28
Replace `log` with optional `tracing`

finnbear:
`tracing`'s output is conditional on a `tracing` subscriber or using the `log` feature of `tracing`. As a result, `tracing` macros can be used in more places, without fear of creating unwanted clutter. Making `log` optional is not as good, since that is a compile-time decision and tracing subscribers can be adjusted at runtime.

Note: optional is a compile time optimization, as tracing statements are inherently optional too.

## Alternative

If logging isn't considered useful for end users, it could be made a dev-dependency and log statements could be conditioned on `#[cfg(test)]`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ license = "MIT"
 
 [dependencies]
 approx = "0.5"
-log = "0.4"
+tracing = { optional = true, version = "0.1" }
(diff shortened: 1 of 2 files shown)


svenstaro:
We use `log` so little that we might as well cut it, I think. If we do want it, I think we should add it back properly.

PR #166, opened 2025-08-03
feat: Add support for no_std

oschijns:
Hello, I made a first draft for adding no_std support for this crate.
I needed to add a few dependencies to replace some std types with no_std equivalents.
I tried to find a way to replace the `thread_local!{BUCKETS}` with something but I am unsure if what I did makes sense...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,11 @@ num-traits = "0.2.19"
 nalgebra = { version = "0.33.0", features = ["default", "serde-serialize"] }
 rayon = { optional = true, version = "1.8.1" }
 wide = { optional = true, version = "0.7.32" }
+hashbrown = { version = "0.15.4" }
+once_cell = { version = "1.21.3", features = [
+    "alloc",
+], default-features = false, optional = true }
+spin = { version = "0.10.0", optional = true }
 
 [dev-dependencies]
 num = "0.4.3"
@@ -31,7 +36,9 @@ float_eq = "1"
 doc-comment = "0.3"
 
 [features]
-default = ["rayon"]
+default = ["std", "rayon"]
+std = []
+no_std = ["dep:once_cell", "dep:spin"]
 bench = []
 simd = ["wide"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ name = "bvh"
 description = "A fast BVH using SAH"
 version = "0.11.0"
 edition = "2021"
-rust-version = "1.65" # may raise to versions that are at least 1y old.
+rust-version = "1.80" # may raise to versions that are at least 1y old.
 authors = [
     "Sven-Hendrik Haase <svenstaro@gmail.com>",
     "Alexander Dmitriev <alexander.dmitriev2580@gmail.com>",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,10 +22,6 @@ nalgebra = { version = "0.33.0", features = ["default", "serde-serialize"] }
 rayon = { optional = true, version = "1.8.1" }
 wide = { optional = true, version = "0.7.32" }
 hashbrown = { version = "0.15.4" }
-once_cell = { version = "1.21.3", features = [
-    "alloc",
-], default-features = false, optional = true }
-spin = { version = "0.10.0", optional = true }
 
 [dev-dependencies]
 num = "0.4.3"
@@ -38,7 +34,6 @@ doc-comment = "0.3"
 [features]
 default = ["std", "rayon"]
 std = []
-no_std = ["dep:once_cell", "dep:spin"]
 bench = []
 simd = ["wide"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,6 @@ num-traits = "0.2.19"
 nalgebra = { version = "0.33.0", features = ["default", "serde-serialize"] }
 rayon = { optional = true, version = "1.8.1" }
 wide = { optional = true, version = "0.7.32" }
-hashbrown = { version = "0.15.4" }
 
 [dev-dependencies]
 num = "0.4.3"
--- a/README.md
+++ b/README.md
@@ -20,6 +20,8 @@ use in a shader this module also exports a flattening procedure, which allows fo
 iterative traversal of the BVH.
 
 It uses `rayon` by default to parallelize building the BVH.
+It can also be used in `no_std` contexts for micro-controllers and retro-consoles
+by disabling default features.
 
 ## Example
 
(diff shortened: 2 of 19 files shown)


src/bvh/bvh_node.rs
```diff
@@ -2,17 +2,21 @@ use crate::aabb::{Aabb, Bounded, IntersectsAabb};
...
+#[cfg(feature = "std")]
 thread_local! {
     /// Thread local for the buckets used while building to reduce allocations during build
-    static BUCKETS: RefCell<[Vec<ShapeIndex>; NUM_BUCKETS]> = RefCell::new(Default::default());
+    static BUCKETS: core::cell::RefCell<[Vec<ShapeIndex>; NUM_BUCKETS]> = core::cell::RefCell::new(Default::default());
 }
 
+#[cfg(not(feature = "std"))]
+static BUCKETS: once_cell::race::OnceBox<spin::Mutex<[Vec<ShapeIndex>; NUM_BUCKETS]>> =
+    once_cell::race::OnceBox::new();
```
finnbear, line 18: Perhaps `no_std` should devolve to allocating `BUCKETS` on the stack/heap? While this adds contention in the memory allocator, using a `spin::Mutex` adds even more contention throughout the entire algorithm.

finnbear (changes requested):
Thanks for the PR! I think `no_std` would be great for `bvh` :)

src/bvh/bvh_node.rs
oschijns: I tried to use an `AtomicBool` but I could not figure out how to guard access to the buckets buffer without using an `Arc`. But then `Arc` doesn't have any `const` initializer. So I got confused how to properly allow multi-thread access to the buckets.
I kind of gave up and wrote a function that allocate new buckets when called.

src/bvh/bvh_node.rs
```diff
@@ -133,14 +122,39 @@ impl<T: BHValue, const D: usize> BvhNode<T, D> {
...
+                BvhNode::build_buckets(
+                    bucket_assignments,
+                    shapes,
+                    indices,
+                    split_axis,
+                    split_axis_size,
+                    &centroid_bounds,
+                    &aabb_bounds,
+                )
+            }
```
finnbear, line 157: Thanks for the changes. To reduce code duplication, you could make a `with_buckets(impl FnMut(&mut BucketArray))` function that, depending on features, either uses `BUCKETS.with` or passes a mutable reference to newly-allocated buckets. Then this code could always use `with_buckets`.

src/bvh/bvh_node.rs
finnbear: If you end up using the `with_buckets` approach I suggested in my latest comment, that function could handle borrowing from `static mut` in a sound and non-`Arc`-requiring way, considering that the borrow would only last for the duration of `closure`.

src/bvh/bvh_node.rs
finnbear: Update: I deduplicated the code. Turns out the closure must be `FnOnce` not `FnMut` because it captures the `indices` variable from the enclosing scope and ends up returning borrows derived from it.

Next steps: I'll re-review your PR within the next few days :)

oschijns: Done.
Yeap, it is easier to read that way.

oschijns:
I didn't check the CI, I will submit fixes this evening.

finnbear (changes requested):
Thank you! Apologies for taking longer than I said I would take for review.

Cargo.toml
```diff
@@ -21,6 +21,7 @@ num-traits = "0.2.19"
 nalgebra = { version = "0.33.0", features = ["default", "serde-serialize"] }
 rayon = { optional = true, version = "1.8.1" }
 wide = { optional = true, version = "0.7.32" }
+hashbrown = { version = "0.15.4" }
```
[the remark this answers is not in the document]
oschijns: According to the Rust doc, the standard library actually use the `HashMap` and `HashSet` implementation provided by `hashbrown`. Maybe I could make it an optional dependency and through some feature flag tricks map `HashMap` and `HashSet` to either the `std` crate or the `hashbrown` crate. I did not considered worth it initially.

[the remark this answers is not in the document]
finnbear: Are you suggesting you used `hashbrown` outside of tests, meaning it must be a `dependency`? I didn't notice any such use. Adding a `dependency` seems like a bigger deal than adding a `dev-dependency`.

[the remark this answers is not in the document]
oschijns: Yes, you are right.
The only places where HashMap and HashSet appear are test modules.
So I will switch `hashbrown` into a `dev-dependency`.
Maybe this dependency is not even necessary at all?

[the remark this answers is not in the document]
oschijns: Since it is only used in tests, I removed the `hashbrown` dependency entirely.
Unit tests cannot be run as is without an `std` environment or without using a third party crate such as `defmt-test`.

finnbear:
Code changes look great! Here's what remains, in my opinion:
- Fixing whatever clippy found ;)
- Rebasing into a few, meaningful, logical commits. There should be a bias towards avoiding commits that undo previous commits (e.g. the temporary inclusion of `hashbrown` shouldn't be reflected in the final commits). It seems easiest, for this PR, to rebase all code changes into a single commit. Once done, request svenstaro's review.
- Writing a changelog entry, which can be a separate commit.

Thanks!

oschijns:
Since Rust 1.80 has been released more than a year ago, I guess it is fine to bump the version? It fixes the dependency resolution.

svenstaro:
@finnbear do you wanna merge this?

svenstaro (changes requested):

svenstaro:
Sorry about the conflicts. Should be easy to fix though.

README.md
```diff
@@ -208,10 +212,12 @@ This will also lead to cases where the BVH would have to be restructured complet
...
 These intersection tests are grouped by dataset and by the BVH generation method.
-* `_after_optimize` uses a BVH which was kept up to date with calls to `optimize`, while
-* `_with_rebuild` uses the same triangle data as `_after_optimize`, but constructs a BVH from scratch.
 
-*120K Triangles*
+- `_after_optimize` uses a BVH which was kept up to date with calls to `optimize`, while
+- `_with_rebuild` uses the same triangle data as `_after_optimize`, but constructs a BVH from scratch.
+
+_120K Triangles_
+
```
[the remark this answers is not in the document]
oschijns: I used zed which by default automatically format files on save.
Sorry about that.

svenstaro (changes requested):
Almost there! Just tiny fixes.

svenstaro:
LGTM! @finnbear any last comments?

svenstaro:
Cool! Thanks @oschijns

oschijns:
Thank you for the crate and for the merge!
:)

